PR #154, opened 2015-01-01
Move Integrations into their own submodules

f2prateek:


f2prateek:
Working towards having moving all integrations under a folder `analytics-integrations` and share common build logic.

PR #203, opened 2015-02-27
Connection Factory

f2prateek:


f2prateek:
Reverting since we don't want this in 3.0.1

PR #208, opened 2015-03-01
Wear module will declare provided dependency on core

f2prateek:


analytics-sample/build.gradle
```diff
@@ -5,8 +5,6 @@ apply from: rootProject.file('gradle/android.gradle')
 
 dependencies {
   compile project(':analytics')
-
-  // only required for Android Wear apps
```
f2prateek: Probably should leave this comment in.

PR #226, opened 2015-03-09
Dont send Mixpanel null values. Closes #224

### LINKED ISSUE #224: Don't send "null" to mixpanel
Hello,

When there are no firstName/lastName or other mixpanel special properties in my identify call, they get sent as "null" to mixpanel.
In the browser or in iOS they are not sent at all, which is expected
For example:
screenshot 2015-03-09 18 26 17
Cheers!

f2prateek:


f2prateek:
Closing this.

PR #235, opened 2015-03-27
Log plan

### LINKED ISSUE #234: Log when an event is disabled in the tracking plan

f2prateek:
Closes #234


f2prateek:
This currently doesn't distinguish between whether the event was disabled in the plan or if the integration simply doesn't support the method.

Also needs tests to verify that the return value of the implemented methods is correct for integrations.

f2prateek:
Will reopen after #247 is merged.

PR #264, opened 2015-04-25
Update Kahuna Integration

TorreyKahuna:
1. Update Kahuna SDK to 556
2. Improved handling of identify and track methods.
3. Added support for screen, alias and reset methods.
4. Added support for e-commerce events via track.


analytics-integrations/kahuna/build.gradle
```diff
@@ -2,5 +2,5 @@ apply from: rootProject.file('gradle/integration.gradle')
 
 dependencies {
   compile rootProject.ext.supportLib
-  compile files('libs/Kahuna-550.jar')
```
f2prateek: Let's delete the old 550 version JAR.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -57,25 +84,64 @@
...
-    KahunaAnalytics.setUserCredential(TWITTER_KEY, identify.traits().getString("twitter"));
-    KahunaAnalytics.setUserCredential(LINKEDIN_KEY, identify.traits().getString("linkedin"));
-
-    KahunaAnalytics.setUserAttributes(identify.traits().toStringMap());
+    KahunaAnalytics.setUserCredential(USER_ID_KEY, identify.userId());
+
+    Traits identityTraits = identify.traits();
+    Map<String, String> userAttributes = new HashMap<>();
+    for (String key : identityTraits.keySet()) {
+      if (identityTraits.get(key) == null) {
```
f2prateek: No need to check for `null` here, we automatically ignore those.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -57,25 +84,64 @@
...
+    KahunaAnalytics.setUserAttributes(userAttributes);
   }
 
   @Override public void track(TrackPayload track) {
     super.track(track);
 
-    // Although not documented, Kahuna wants revenue in cents
-    KahunaAnalytics.trackEvent(track.event(), track.properties().getInt("quantity", 0),
-        (int) (track.properties().revenue() * 100));
+    if (track == null || TextUtils.isEmpty(track.event()))
```
f2prateek: No need to check for these, neither of them can ever be true.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -57,25 +84,64 @@
...
+      return;
+
+    String kahunaEventName = track.event();
+
+    kahunaEcommerceHelper.process(track);
+
+    if (track.properties() == null) {
```
f2prateek: Properties can never be null, they can be empty though so you can check for the size instead (or wrap it in `isNullOrEmpty`).

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -57,25 +84,64 @@
...
+      KahunaAnalytics.trackEvent(kahunaEventName);
+    } else {
+      Object revenueObject = track.properties().get(SEGMENT_REVENUE_KEY);
```
f2prateek: Delete SEGMENT_REVENUE_KEY and replace this line with `int revenue = (int) (track.properties().revenue() * 100)`

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -57,25 +84,64 @@
...
+      Object quantityObject = track.properties().get(SEGMENT_QUANTITY_KEY);
+      if (revenueObject == null && quantityObject == null) {
+        KahunaAnalytics.trackEvent(kahunaEventName);
+      } else {
+        int kahunaValue = (revenueObject == null) ? 0 : (int) (track.properties().revenue() * 100);
```
f2prateek: Document that Kahuna accepts revenue in cents and not dollars.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -57,25 +84,64 @@
...
+        int kahunaCount = (quantityObject == null) ? 0
+                : track.properties().getInt(SEGMENT_QUANTITY_KEY, 0);
+        KahunaAnalytics.trackEvent(kahunaEventName, kahunaCount, kahunaValue);
```
f2prateek: Doesn't seem to be that this whole block is needed. The earlier approach `KahunaAnalytics.trackEvent(track.event(), track.properties().getInt("quantity", 0), (int) (track.properties().revenue() * 100));` is a one liner and simpler to read (and can be wrapped in a conditional to check if the revenue and quantity are defined).

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -57,25 +84,64 @@
...
+      }
+    }
+  }
+
+  @Override public void screen(ScreenPayload screen) {
+    super.screen(screen);
+
+    // TODO: Segment is working on the trackAllPages setting, currently it is defaulted to false.
+    if (trackAllPages) {
+      if (screen != null && screen.name() != null) {
```
f2prateek: No need to check for these, both will never be true.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -83,4 +149,14 @@
 
     KahunaAnalytics.logout();
   }
+
+  @Override public void alias(AliasPayload alias) {
+    KahunaAnalytics.setUserCredential(USER_ID_KEY, alias.userId());
```
f2prateek: This doesn't look right - if you don't support alias don't implement this method. Users are instructed to call `identify` after calling alias so you'll get the new user id anyway. `Alias` is reserved when you explicitly want to link two IDs, previousId and currentId together.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -83,4 +149,14 @@
...
+  }
+
+  private static String getISO8601StringForDate(Date date) {
```
f2prateek: You can reuse this method - https://github.com/segmentio/analytics-android/blob/master/analytics-core/src/main/java/com/segment/analytics/internal/Utils.java#L71 instead.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaEcommerceHelper.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Set;
+
+/**
+ * Kahuna helps mobile marketers send push notifications and in-app messages.
+ *
+ * @see <a href="https://www.kahuna.com/">Kahuna</a>
+ * @see <a href="https://segment.com/docs/integrations/kahuna/">Kahuna Integration</a>
+ * @see <a href="http://app.usekahuna.com/tap/getstarted/android/">Kahuna Android SDK</a>
+ */
+public class KahunaEcommerceHelper {
```
f2prateek: Make this class package protected.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaEcommerceHelper.java
```diff
@@ -0,0 +1,134 @@
...
+import com.segment.analytics.internal.model.payloads.TrackPayload;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Kahuna helps mobile marketers send push notifications and in-app messages.
```
f2prateek: We can probably remove these docs, they're the same as the ones on KahunaIntegration.java

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaEcommerceHelper.java
```diff
@@ -0,0 +1,134 @@
...
+  private static final String LAST_PURCHASE_DISCOUNT = "Last Purchase Discount";
+
+  private static final int MAX_CATEGORIES_VIEWED_ENTRIES = 50;
+  private static final String NONE = "None";
+
+  public void process(TrackPayload track) {
+    if (track != null && !TextUtils.isEmpty(track.event()) && track.properties() != null) {
+
+      if (SEG_VIEWED_PRODUCT_CATEGORY.equalsIgnoreCase(track.event())) {
+
```
f2prateek: Remove these whitespaces.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -29,14 +47,23 @@
...
   @Override public void initialize(Context context, ValueMap settings, LogLevel logLevel)
       throws IllegalStateException {
     if (!isOnClassPath("android.support.v4.app.Fragment")) {
       throw new IllegalStateException("Kahuna requires the support library to be bundled.");
     }
+    trackAllPages = settings.getBoolean("trackAllPages", false);
+
     KahunaAnalytics.onAppCreate(context, settings.getString("apiKey"),
-        settings.getString("pushSenderId"));
+             settings.getString("pushSenderId"));
```
f2prateek: unnecessary tabs added here.

f2prateek:
Thanks, I've made a few comments inline. In particular, I don't think we should implement `alias`, at least not this way. Right now it's simply storing the userId, but you'll get that in the next call to identify anyway (which happens right after since this how ask users to implement it in their code https://github.com/segmentio/analytics-android/blob/master/analytics-core/src/main/java/com/segment/analytics/Analytics.java#L370).

I haven't taken a detailed look at the KahunaHelper class yet, but we'd also need to add tests for the KahunaHelper class, and squash rebase these commits against master since we like having each individual commit pass our build for git bisect.

In addition to those, we'll have to reformat some of the code before we can merge this, I'm pretty sure check style will fail the build otherwise. Our CI is also disabled for 3rd party PRs, so please make sure you've run the build locally with `make build` or `./gradlew build`.

 Let me know if you'd like to finish it up or I can take it from here next week.

TorreyKahuna:
Thanks, I'll have my team go through and make some fixes based on your feedback.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
rayleeriver: this is added for defensive purpose.  Prefer to leave the check in the code.

TorreyKahuna:
@f2prateek, We've made some changes based on your comments and added in some more tests.

Can you make the changes necessary to get it working nicely in your build system? Thanks!

f2prateek:
@TorreyKahuna thanks, we're actually making some internal changes for roots integrations that will affect this PR. I'll take care of updating this PR with those new APIs, and I'll ping you when it's done for a quick review to make sure I didn't miss anything.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaEcommerceHelper.java
```diff
@@ -0,0 +1,115 @@
...
+      kahunaCategoriesSet.add(track.properties().category());
+      Map<String, String> userAttributes = KahunaAnalytics.getUserAttributes();
+      userAttributes.put(CATEGORIES_VIEWED, TextUtils.join(",", kahunaCategoriesSet));
+      userAttributes.put(LAST_VIEWED_CATEGORY, track.properties().category());
+      KahunaAnalytics.setUserAttributes(userAttributes);
+    } else {
+      Map<String, String> userAttributes = KahunaAnalytics.getUserAttributes();
+      userAttributes.put(LAST_VIEWED_CATEGORY, NONE);
+      if (userAttributes.get(CATEGORIES_VIEWED) == null) {
+        userAttributes.put(CATEGORIES_VIEWED, NONE);
```
f2prateek: Why do you need a category `NONE` - don't you already have that if a category has not been previously set?

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaEcommerceHelper.java
```diff
@@ -0,0 +1,115 @@
...
+    if (!TextUtils.isEmpty(track.properties().category())) {
+      Set<String> kahunaCategoriesSet
+              = Collections.newSetFromMap(new LinkedHashMap<String, Boolean>() {
+        protected boolean removeEldestEntry(Entry<String, Boolean> eldest) {
+          return size() > MAX_CATEGORIES_VIEWED_ENTRIES;
+        }
+      });
+      if (KahunaAnalytics.getUserAttributes().containsKey(CATEGORIES_VIEWED)) {
+        String serializedCategories = KahunaAnalytics.getUserAttributes().get(CATEGORIES_VIEWED);
+        kahunaCategoriesSet.addAll(Arrays.asList(serializedCategories.split(",")));
```
f2prateek: So if there is a track event without a category, then CATEGORIES_VIEWED = None (from the line below). Then if there is another track event with a category (say sports), because this joins the categories, it would be "NONE, Sports" which seems weird.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaEcommerceHelper.java
f2prateek: Also what happens for users with multiple devices? Say I have an Android phone and an Android tablet. I browse on my Android tablet, view a bunch of categories that is sent to Kahuna. Then I switch to my phone, where I actually don't view any categories - so no it's NONE and that's what it shows up in Kahuna as. Is that the expected behaviour?

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaEcommerceHelper.java
```diff
@@ -0,0 +1,115 @@
...
+  // Kahuna constants
+  private static final String CATEGORIES_VIEWED = "Categories Viewed";
+  private static final String LAST_VIEWED_CATEGORY = "Last Viewed Category";
+  private static final String LAST_PRODUCT_VIEWED_NAME = "Last Product Viewed Name";
+  private static final String LAST_PRODUCT_VIEWED_ID = "Last Product Viewed Id";
+  private static final String LAST_PRODUCT_ADDED_TO_CART_NAME = "Last Product Added To Cart Name";
+  private static final String LAST_PRODUCT_ADDED_TO_CART_CATEGORY
+          = "Last Product Added To Cart Category";
+  private static final String LAST_PURCHASE_DISCOUNT = "Last Purchase Discount";
+  private static final int MAX_CATEGORIES_VIEWED_ENTRIES = 50;
```
f2prateek: Is there a reason to limit to 50 categories?

f2prateek: I might be wrong here though since I'm not too familiar with how Kahuna works under the hood.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -57,25 +81,64 @@
...
+        // and make sure to format any Date objects appropriately.
+        Object value = identityTraits.get(key);
+        if (value instanceof Date) {
+          userAttributes.put(key, Utils.toISO8601Date((Date) value));
+        } else {
+          userAttributes.put(key, String.valueOf(value));
+        }
+      }
+    }
+    KahunaAnalytics.setUserAttributes(userAttributes);
```
f2prateek: In track, we've done `Map<String, String> userAttributes = KahunaAnalytics.getUserAttributes();` and then added mappings to that. Should it be the same here - since it looks like this might override existing userAttributes.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaEcommerceHelper.java
TorreyKahuna: We track the category NONE because it is an optional field when tracking something like Last Product Viewed and we don't want our attributes to be out of sync by not tracking a last category correctly.

To answer your 2nd question, yes. Kahuna tracks Users and not devices, so from a user perspective, your last category viewed in your profile would correspond to the last category you actually looked at, regardless of phone, tablet or web. We handle merging as needed on our server side.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaEcommerceHelper.java
TorreyKahuna: Great feedback. We debated on this internally and since we need to track a last category viewed of None if it isn't passed in by the developer, we need to track in the list of Categories Viewed as well to remain consistent. We think that while this is weird, it would reflect the correct state of the customer's information if they are tracking something without a category we'll reflect as None.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaEcommerceHelper.java
TorreyKahuna: We limit to 50 because we don't wanna let this grow unbounded client side and anything more than 50 starts becoming not as useful from a Kahuna campaign perspective. For this initial phase of the integration, we decided to track 50 unique categories while keeping in mind to not make things too complicated as we are also going to release a newer SDK soon with newer APIs that will make future segment iterations much better. Great question though!

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
TorreyKahuna: For the categories_viewed attribute, we add to the existing list that the SDK knows (a comma separated list). In here, we are tracking new keys (or overriding the values those key reference) and our Mobile SDK takes care of merging the attributes internally and letting the server know of any delta changes as needed. For these attributes we don't care about the previous values, only the new ones so we don't need to get the previous values.

TorreyKahuna:
@f2prateek Great questions, I've responded back to them with our reasoning for implementing with that approach. Please let me know when your newer PR is ready for me to take a look. Thanks!

analytics-integrations/kahuna/src/test/java/com/segment/analytics/internal/integrations/KahunaTest.java
```diff
@@ -49,6 +58,26 @@
...
+  private static final String SEG_COMPLETED_ORDER = "Completed Order";
+  private static final String SEG_DISCOUNT_KEY = "discount";
+  private static final String SEG_NAME_KEY = "name";
+  private static final String SEG_CATEGORY_KEY = "category";
+
+  // Kahuna constants
+  private static final String CATEGORIES_VIEWED = "Categories Viewed";
+  private static final String LAST_VIEWED_CATEGORY = "Last Viewed Category";
+  private static final String LAST_PRODUCT_VIEWED_NAME = "Last Product Viewed Name";
+  private static final String LAST_PRODUCT_VIEWED_ID = "Last Product Viewed Id";
```
f2prateek: This seems to be unused?

PR #265, opened 2015-04-27
Fix NPE on googleAnalyticsInstance.newTracker(...)

halfhp:
Swapped the order of googleAnalyticsInstance and tracker initialization to avoid the NPE.


f2prateek:
Thanks!

PR #272, opened 2015-05-04
Update kahuna integration

f2prateek:


analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -27,8 +39,24 @@
  * @see <a href="http://app.usekahuna.com/tap/getstarted/android/">Kahuna Android SDK</a>
  */
 public class KahunaIntegration extends AbstractIntegration<Void> {
+  private static final String CATEGORIES_VIEWED = "Categories Viewed";
+  private static final String LAST_VIEWED_CATEGORY = "Last Viewed Category";
+  private static final String LAST_PRODUCT_VIEWED_NAME = "Last Product Viewed Name";
+  private static final String LAST_PRODUCT_VIEWED_ID = "Last Product Viewed Id";
+  private static final String LAST_PRODUCT_ADDED_TO_CART_NAME = "Last Product Added To Cart Name";
+  private static final String LAST_PRODUCT_ADDED_TO_CART_CATEGORY =
```
f2prateek: This seems to be unused?

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -27,8 +39,24 @@
  * @see <a href="http://app.usekahuna.com/tap/getstarted/android/">Kahuna Android SDK</a>
  */
 public class KahunaIntegration extends AbstractIntegration<Void> {
+  private static final String CATEGORIES_VIEWED = "Categories Viewed";
+  private static final String LAST_VIEWED_CATEGORY = "Last Viewed Category";
+  private static final String LAST_PRODUCT_VIEWED_NAME = "Last Product Viewed Name";
+  private static final String LAST_PRODUCT_VIEWED_ID = "Last Product Viewed Id";
```
f2prateek: This seems to be unused.

f2prateek:
@TorreyKahuna mind taking one last look at this? I removed `LAST_PRODUCT_VIEWED_ID` since it was unused.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -60,25 +89,119 @@
...
-        identify.traits().email());
-
-    KahunaAnalytics.setUserCredential(USERNAME_KEY, identify.traits().username());
-    KahunaAnalytics.setUserCredential(EMAIL_KEY, identify.traits().email());
-    KahunaAnalytics.setUserCredential(FACEBOOK_KEY, identify.traits().getString("facebook"));
-    KahunaAnalytics.setUserCredential(TWITTER_KEY, identify.traits().getString("twitter"));
-    KahunaAnalytics.setUserCredential(LINKEDIN_KEY, identify.traits().getString("linkedin"));
-
-    KahunaAnalytics.setUserAttributes(identify.traits().toStringMap());
+    Traits traits = identify.traits();
```
TorreyKahuna: It looks like we are missing a call here, we should be tracking the userID also with the following:

KahunaAnalytics.setUserCredential(USER_ID_KEY, identify.userId());

Kahuna tracks UserId as "user_id" and Segment tracks as "userId" so here we track it explicitly outside of the traits for loop.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -60,25 +89,119 @@
...
+      trackViewedProductCategory(track);
+    } else if (VIEWED_PRODUCT.equalsIgnoreCase(event)) {
+      trackViewedProduct(track);
+      trackViewedProductCategory(track);
+    } else if (ADDED_PRODUCT.equalsIgnoreCase(event)) {
+      trackAddedProduct(track);
+      trackAddedProductCategory(track);
+    } else if (COMPLETED_ORDER.equalsIgnoreCase(event)) {
+      trackCompletedOrder(track);
+    }
```
TorreyKahuna: Can we move this logic up above track event? We prefer tracking the attributes before tracking the event.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -36,6 +64,7 @@
       throw new IllegalStateException("Kahuna requires the support library to be bundled.");
     }
 
+    trackAllPages = settings.getBoolean("trackAllPages", true);
```
rayleeriver: Why trackAllPages default to true if not set by user?  This may cause a lot of unintended screen events.

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
```diff
@@ -60,25 +89,119 @@
...
+    Map<String, String> userAttributes = KahunaAnalytics.getUserAttributes();
+    userAttributes.put(LAST_PURCHASE_DISCOUNT, String.valueOf(discount));
+    KahunaAnalytics.setUserAttributes(userAttributes);
+  }
+
+  @Override public void screen(ScreenPayload screen) {
+    super.screen(screen);
+
+    if (trackAllPages) {
+      KahunaAnalytics.trackEvent(String.format(VIEWED_EVENT_FORMAT, screen.event()));
```
rayleeriver: Looks like screen.event will track screen.category() as the screen's name if the screen.name() is not set.  Not sure if this is what we want.

TorreyKahuna:
@f2prateek We made a few comments, please have a look. Thanks!

analytics-integrations/kahuna/src/main/java/com/segment/analytics/internal/integrations/KahunaIntegration.java
f2prateek: Screen names are optional, for a sports category, this would become `Viewed Sports Page`, which is better than swallowing the event.

Alternatively Amplitude (see `screen` method in `AmplitudeIntegration.java`) has a setting to track named pages and to track categorized pages only. In practice however, we've found it's not useful to users.

f2prateek:
@TorreyKahuna made the changes.

TorreyKahuna:
@f2prateek the changes look good from our side. Thanks!

f2prateek:
Awesome, thanks for you help!

PR #282, opened 2015-05-19
Add Taplytics Integration

f2prateek:


analytics-integrations/taplytics/src/main/java/com/segment/analytics/internal/integrations/TaplyticsIntegration.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Override public String key() {
+    return TAPLYTICS_KEY;
+  }
+
+  @Override public void track(TrackPayload track) {
+    super.track(track);
+    JSONObject metaData = track.properties().toJsonObject();
+    Taplytics.logEvent(track.event(), track.properties().value(), metaData);
```
f2prateek: @VicV we have something called `value` in our properties that should be the same as your generic Number value associated with an event. I've updated the integration to look for that.

VicV: @f2prateek is that guaranteed to be a Number?

f2prateek: The method is an alias to `getDouble()`, if it isn't a number then it returns zero.

VicV: Perfect

analytics-integrations/taplytics/src/main/java/com/segment/analytics/internal/integrations/TaplyticsIntegration.java
```diff
@@ -0,0 +1,98 @@
...
+    } catch (JSONException ignored) {
+    }
+  }
+
+  @Override public void group(GroupPayload group) {
+    super.group(group);
+
+    JSONObject customData = group.traits().toJsonObject();
+    JSONObject userAttributes = new JSONObject();
+    insert(userAttributes, "customData", customData);
```
f2prateek: @VicV I'm a bit unsure about adding group traits to the `customData` field, which is already being used for user traits in identify. I'd like to test this out and see how this looks on the Taplytics dashboard so we can figure the best way to send this to Taplytics.

analytics-integrations/taplytics/build.gradle
```diff
@@ -0,0 +1,5 @@
+apply from: rootProject.file('gradle/integration.gradle')
+
+dependencies {
+  compile files('libs/taplytics-1.4.2.jar')
```
f2prateek: Does this JAR contain all the relevant dependencies listed here - https://cloudup.com/c6GLI7ewTHR

VicV: Yep.

analytics-integrations/taplytics/src/main/java/com/segment/analytics/internal/integrations/TaplyticsIntegration.java
VicV: We don't even have a 'groups' trait. Which is why I would just like to collect it as custom data.

analytics-integrations/taplytics/build.gradle
f2prateek: cc @VicV

PR #285, opened 2015-05-20
Taplytics Integration

VicV:
Hey @f2prateek, because of the formatting differences everything broke when trying to create a PR into pr/281.

Essentially, since 1.4.3, I changed the way user attributes work. We just take all data and sort it out on our side as people who use Taplytics would expect it to be sorted out.

As for the 'groups', I'm just taking whatever 'groups' is and putting it into our attributes. 

This is a branch off of Pr/281 so it has all of your changes already.


VicV:
@f2prateek actually I fixed up groups.

VicV:
cc @f2prateek

PR #290, opened 2015-06-01
Add Taplytics Integration

f2prateek:
Closes #285


analytics-integrations/taplytics/build.gradle
```diff
@@ -0,0 +1,5 @@
+apply from: rootProject.file('gradle/integration.gradle')
+
+dependencies {
+  compile files('libs/Taplytics-1.4.3.jar')
```
f2prateek: @VicV Do you have a version of the JAR that doesn't include all your dependencies? We have other libraries that depend on OkHttp for instance, and not using a package manager will cause conflicts.

VicV: @f2prateek no, that dependency would be a gradle dependency then and can be switched to the following:

1. _In your module’s build.gradle, add the url to the sdk._

	  ```
  repositories {                                                                                              
    maven { url "https://github.com/taplytics/Taplytics-Android-SDK/raw/master/AndroidStudio/" }
  }      
  ```
  
2. _In your module’s build.gradle dependencies, compile Taplytics and its dependencies._

  	```
  dependencies {                                                                   
    //Taplytics                                                                        
    compile("com.taplytics.sdk:taplytics:+@aar")  
    
    //Dependencies for taplytics
    compile("com.mcxiaoke.volley:library:+")
    compile("com.squareup.okhttp:okhttp-urlconnection:+")
    compile("com.squareup.okhttp:okhttp:+")
   
    //Excluding org.json due to compiler warnings
    //socket.io connections only made on debug devices OR if making live changes to a release build.
    //No socket.io connection will be made on your release devices unless explicitly told to do so. 
    compile("com.github.nkzawa:socket.io-client:+") {
        exclude group: 'org.json'
    }
    compile("com.github.nkzawa:engine.io-client:+") {
        exclude group: 'org.json'
    }
    
    //Only include this if you wish to enable push notifications:
    compile("com.google.android.gms:play-services-gcm:7.5.0")
  }    
  ```
  
It is worth noting that the dependencies in that jar are all from my special forks which are extremely cut down versions of the respective libs. They are also proguarded down to animal names so a conflict is extremely unlikely.

I wasn't sure how segment worked with gradle dependencies.

PR #298, opened 2015-06-12
Taplytics Version Up

VicV:


f2prateek:
You'll have to update the reference in the `build.gradle` file as well  - https://github.com/segmentio/analytics-android/blob/master/analytics-integrations/taplytics/build.gradle#L4.

Also please leave the jar name starting with caps `Taplytics-...` instead of `taplytics-...`

f2prateek:
You can run `./gradlew build` from the command line to verify everything builds as well.

VicV:
righto

VicV:
All set @f2prateek

PR #303, opened 2015-06-22
track revenue in localytics

### LINKED ISSUE #301: Localytics: 'Track' does not recognize 'revenue' property correctly on Android
On the iOS version on Segment, the track event checks to see if there is a 'revenue' present
https://github.com/segmentio/analytics-ios/blob/master/Analytics/Integrations/Localytics/SEGLocalyticsIntegration.m#L127-L133

On the Android version, it doesn't:
https://github.com/segmentio/analytics-android/blob/21161a10e8b8b08f84c6331b80af005e766a0467/analytics-integrations/localytics/src/main/java/com/segment/analytics/internal/integrations/LocalyticsIntegration.java#L127-L133

Without this, it's not possible to track monetary events in Localytics.

nashira:
this fixes #301

### DIFF
--- a/analytics-integrations/localytics/src/main/java/com/segment/analytics/internal/integrations/LocalyticsIntegration.java
+++ b/analytics-integrations/localytics/src/main/java/com/segment/analytics/internal/integrations/LocalyticsIntegration.java
@@ -128,7 +128,14 @@ public class LocalyticsIntegration extends AbstractIntegration<Void> {
     super.track(track);
     Properties props = track.properties();
     setContext(track.context());
-    Localytics.tagEvent(track.event(), props.toStringMap());
+
+    final long revenue = (long) (props.revenue() * 100);
+    if (revenue != 0) {
+      Localytics.tagEvent(track.event(), props.toStringMap(), revenue);
+    } else {
+      Localytics.tagEvent(track.event(), props.toStringMap());
+    }
+
     setCustomDimensions(props);
   }
 


analytics-integrations/localytics/src/main/java/com/segment/analytics/internal/integrations/LocalyticsIntegration.java
```diff
@@ -128,7 +128,14 @@
     super.track(track);
     Properties props = track.properties();
     setContext(track.context());
-    Localytics.tagEvent(track.event(), props.toStringMap());
+
+    final long revenue = (long) (props.revenue() * 100);
```
f2prateek: Could we move the multiplication inside the `if` block?

```
   double revenue = props.revenue();
    if (revenue != 0.00) {
      Localytics.tagEvent(track.event(), props.toStringMap(), (long) revenue * 100);
    } 
```

f2prateek:
Thanks, also needs a test case to make sure we call the correct overloaded method.

analytics-integrations/localytics/src/main/java/com/segment/analytics/internal/integrations/LocalyticsIntegration.java
nashira: I think it's better to check on a long than a double.  For instance, what happens if revenue gets passed in as 0.005, or half a cent?  It will pass the condition but then get truncated and tracked as zero.  Also, it should take about a nanosecond to multiply, you don't save anything you can measure.

f2prateek: Good point, I think if you pass it in as under a cent, it's reasonable for us to just round to 0 cents. Hardly an issue in real code I think though, so I'm fine leaving it as is.

f2prateek:
@nashira noticed you were passing the LTV increases in cents, any reason in particular?

f2prateek:
nvm found the docs here http://docs.localytics.com/index.html#Dev/Instrument/customer-ltv.html

f2prateek:
Closed with #304

PR #307, opened 2015-06-25
Add support for incrementing customer value in Localytics

sperand-io:
not sure if this is the correct way of handling just yet... but this is a first crack at adding support for customer LTV in localytics, a la: http://docs.localytics.com/index.html#Dev/Instrument/customer-ltv.html

iOS parity, for reference: https://github.com/segmentio/analytics-ios/blob/master/Analytics/Integrations/Localytics/SEGLocalyticsIntegration.m#L127-L134


f2prateek:
I think this is a dupe of #303. We just need tests in that one of you want to take that on.

The CI error is from using properties instead of props as the variable name.

sperand-io:
aha! should have caught both of those things. thanks @f2prateek!

f2prateek:
Changed it to use cents

sperand-io:
perfect — thanks!

PR #310, opened 2015-06-30
Retrieves Apptimize as a Gradle dependency instead of a Jar

imminent:
Utilitizing Gradle to retrieve the dependency allows for Gradle transitivity in dependency management.


f2prateek:
I've looked at this in the past and what blocked this was the behaviour for users that don't declare the custom repository block. It's less than ideal if users of the default module have to declare `maven { url 'http://maven.apptimize.com/artifactory/repo' }`.

f2prateek:
Closing until Apptimize is on Maven Central.

PR #311, opened 2015-06-30
Adds ability to set an initial capacity for Properties

imminent:
Setting initial capacity sets the initial capacity of the delegated Map. This saves on unnecessary allocation when the map has a consistent size, which is common for tracking `Properties`.

### DIFF
--- a/analytics-core/src/main/java/com/segment/analytics/ValueMap.java
+++ b/analytics-core/src/main/java/com/segment/analytics/ValueMap.java
@@ -51,6 +51,10 @@ public class ValueMap implements Map<String, Object> {
     delegate = new LinkedHashMap<>();
   }
 
+  public ValueMap(int initialCapacity) {
+    delegate = new LinkedHashMap<>(initialCapacity);
+  }
+
   public ValueMap(Map<String, Object> map) {
     if (map == null) {
       throw new IllegalArgumentException("Map must not be null.");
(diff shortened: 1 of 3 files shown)


f2prateek:
I think a better solution would update our API to accept a regular `Map<String, ?>` as properties, which users can optionally size in advance since they would be able to pass in their own map implementation.

If not, at the very least we should update the other value types (traits) to expose the int overload.

f2prateek:
I guess with the first solution you'd lose the type safety of properties (e.g. `properties.putName(...) vs properties.put("name", ...)`) which is less than ideal. The int overload would work better in that case.

imminent:
If I understand your second approach, you're saying that `Traits` needs the same treatment. I would agree with that.

analytics-core/src/main/java/com/segment/analytics/ValueMap.java
```diff
@@ -51,6 +51,10 @@ public ValueMap() {
     delegate = new LinkedHashMap<>();
   }
 
+  public ValueMap(int initialCapacity) {
```
f2prateek: Is there any value in making this `public` over `protected` or package-private? Callers wouldn't really create instances of this directly. Those who really need the int overload, can create their own pre-sized maps and use the map constructor.

imminent: Since the other constructor is public, it makes sense for this one to be
public as well.

PR #312, opened 2015-06-30
Removes enum types in favor of StringDef and IntDef constants

imminent:
Using `@StringDef` and `@IntDef` still gives type safety without the memory overhead of `enum`s.


f2prateek:
Sorry most of these changes are a breaking API changes that will require a major version bump, which I don't see happening anytime soon.

PR #314, opened 2015-07-07
Incorporates Tracking Plan filtering to event logging

imminent:
The logging currently doesn't properly convey when the event will be filtered from an integration, so it always shows are sending to all Bundled Integrations, even when it is not.

### DIFF
--- a/analytics-core/src/main/java/com/segment/analytics/IntegrationManager.java
+++ b/analytics-core/src/main/java/com/segment/analytics/IntegrationManager.java
@@ -383,7 +383,7 @@ class IntegrationManager implements Application.ActivityLifecycleCallbacks {
   /** Runs the given operation on all bundled integrations. */
   void run(IntegrationOperation operation) {
     if (logLevel.log()) {
-      debug("Running %s on %s integrations.", operation, integrations.size());
+      debug("Running %s.", operation);
     }
     for (int i = 0; i < integrations.size(); i++) {
       AbstractIntegration integration = integrations.get(i);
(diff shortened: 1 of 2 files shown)


f2prateek:
I think it might be nicer to just drop the number of integrations part here. I don't think there's any value in printing the number of integrations an operation is running on.

```java
debug("Running %s.", operation);
```

And then additionally we could tweak the integration [operation interface](https://github.com/segmentio/analytics-android/blob/master/analytics-core/src/main/java/com/segment/analytics/IntegrationOperation.java#L233).
```java
interface IntegrationOperation {
  abstract *boolean* run(AbstractIntegration integration, ProjectSettings projectSettings);
}
```

And so basically it will return false if the event got stripped by the tracking plan, and we can [log that](https://github.com/segmentio/analytics-android/blob/master/analytics-core/src/main/java/com/segment/analytics/IntegrationManager.java#L391).
```java
bool ran = operation.run(integration, projectSettingsCache.get());

if (logLevel.log()) {
  if(ran) {
    debug("Took %s ms to run action %s on %s.", duration, operation, integration.key());
  } else {
    debug("Skipped %s on %s.", operation, integration.key());
  } 
}
```

What do you think?

imminent:
Yes, the first part makes sense, other than mirroring the filter count on
the tracking plan (which is minimally useful). I went with not changing the
signature of `run` because only 1 method `track` does anything different.
So separating it gives more flexibility and no API change at the cost of
being slightly less efficient.

imminent:
I changed it to:
`debug("Running %s.", operation);`

f2prateek:
Closing until we decide on the correct implementation for this.

PR #324, opened 2015-07-29
Makes Properties a null-safe Map

imminent:
This prevents keys in the `Properties` from having null values. Which is an issue I ran into with the Mixpanel SDK and implemented in our code. Wanted to share it in case others could end up in the same problem.


f2prateek:
Do you have the Mixpanel stacktrace by any chance? We can fix in the mapping layer instead of disallowing it everywhere.

PR #333, opened 2015-08-21
Google Analytics: Add Support for Custom Dimensions and Metrics

sperand-io:
- [x] time with @f2prateek to go over and get 20% review
- [x] get it so this isn't breaking any existing tests
- [x] write tests for CDs and CMs
- [x] get those passing 
- [ ] final review by @f2prateek

### DIFF
--- a/analytics-integrations/google-analytics/src/main/java/com/segment/analytics/internal/integrations/GoogleAnalyticsIntegration.java
+++ b/analytics-integrations/google-analytics/src/main/java/com/segment/analytics/internal/integrations/GoogleAnalyticsIntegration.java
@@ -12,6 +12,7 @@ import com.segment.analytics.Analytics;
 import com.segment.analytics.Properties;
 import com.segment.analytics.ValueMap;
 import com.segment.analytics.internal.AbstractIntegration;
+import com.segment.analytics.internal.Utils;
 import com.segment.analytics.internal.model.payloads.IdentifyPayload;
 import com.segment.analytics.internal.model.payloads.ScreenPayload;
 import com.segment.analytics.internal.model.payloads.TrackPayload;
@@ -43,6 +44,8 @@ public class GoogleAnalyticsIntegration extends AbstractIntegration<Tracker> {
   Tracker tracker;
   GoogleAnalytics googleAnalyticsInstance;
   boolean sendUserId;
+  ValueMap customDimensions;
+  ValueMap customMetrics;
 
   @Override public void initialize(Analytics analytics, ValueMap settings)
       throws IllegalStateException {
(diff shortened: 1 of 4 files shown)


analytics-integrations/google-analytics/src/main/java/com/segment/analytics/internal/integrations/GoogleAnalyticsIntegration.java
```diff
@@ -166,6 +202,20 @@ boolean handleProductEvent(String event, String category, Properties properties)
     return false;
   }
 
+  void setCustomDimensions(HitBuilders.HitBuilder hit, Properties properties) {
```
sperand-io: looks like HitBuilders.HitBuilder is protected, so I'm not sure of how to abstract this out, may just have to have some repetitive logic in `.screen` and `.track`

analytics-integrations/google-analytics/src/main/java/com/segment/analytics/internal/integrations/GoogleAnalyticsIntegration.java
```diff
@@ -73,6 +76,16 @@
     }
     // tracker.setSampleRate(settings.getDouble("siteSpeedSampleRate", 1));
     sendUserId = settings.getBoolean("sendUserId", false);
+
+    customDimensions = settings.getValueMap("dimensions");
```
f2prateek: This should already be an empty map coming from the server? No need to have the `null` check below.

analytics-integrations/google-analytics/src/main/java/com/segment/analytics/internal/integrations/GoogleAnalyticsIntegration.java
f2prateek: We could create two new classes `EventHitBuilder` and `AppHitViewBuilder` that extend `EventBuilder` and `AppViewBuilder` respectively, and implement a new interface `CustomHitBuilder` that this method accepts. 

`void setCustomDimensions(CustomHitBuilder hit, Properties properties)`

```java
  interface CustomHitBuilder {
    CustomHitBuilder setCustomDimension(int index, String dimension);
  }

  static class EventHitBuilder extends HitBuilders.EventBuilder implements CustomHitBuilder {
    @Override public EventHitBuilder setCustomDimension(int index, String dimension) {
      super.setCustomDimension(index, dimension);
      return this;
    }
  }

  static class AppViewHitBuilder extends HitBuilders.EventBuilder implements CustomHitBuilder {
    @Override public AppViewHitBuilder setCustomDimension(int index, String dimension) {
      super.setCustomDimension(index, dimension);
      return this;
    }
  }
```

sperand-io: nice! thanks @f2prateek :)

PR #335, opened 2015-08-25
Updating Amplitude SDK to v2.0.2

djih:
Updating Amplitude SDK to v2.0.2.

* API change - removed startSession() and endSession(). Sessions are now handled by attaching an ActivityLifecycleCallback (calling enableForeGroundTracking during init and passing in the Application). See Setup, bullet 4: https://github.com/amplitude/Amplitude-Android#setup
* Start/end session events are no longer logged by default; however, users can re-enable them by calling trackSessionEvents(true) on the Amplitude Client. This is exposed in Segment via the trackSessionEvents setting.

### DIFF
--- a/analytics-integrations/amplitude/src/main/java/com/segment/analytics/internal/integrations/AmplitudeIntegration.java
+++ b/analytics-integrations/amplitude/src/main/java/com/segment/analytics/internal/integrations/AmplitudeIntegration.java
@@ -1,6 +1,5 @@
 package com.segment.analytics.internal.integrations;
 
-import android.app.Activity;
 import com.amplitude.api.AmplitudeClient;
 import com.segment.analytics.Analytics;
 import com.segment.analytics.Properties;
@@ -34,8 +33,13 @@ public class AmplitudeIntegration extends AbstractIntegration<Void> {
     trackAllPages = settings.getBoolean("trackAllPages", false);
     trackCategorizedPages = settings.getBoolean("trackCategorizedPages", false);
     trackNamedPages = settings.getBoolean("trackNamedPages", false);
+    boolean trackSessionEvents = settings.getBoolean("trackSessionEvents", false);
     amplitude = AmplitudeClient.getInstance();
     amplitude.initialize(analytics.getApplication(), settings.getString("apiKey"));
+    amplitude.enableForegroundTracking(analytics.getApplication());
+    if (trackSessionEvents) {
+      amplitude.trackSessionEvents(true);
+    }
   }
 
   @Override public Void getUnderlyingInstance() {
@@ -44,16 +48,6 @@ public class AmplitudeIntegration extends AbstractIntegration<Void> {
 
   @Override public String key() {
     return AMPLITUDE_KEY;
-  }
-
-  @Override public void onActivityResumed(Activity activity) {
-    super.onActivityResumed(activity);
-    amplitude.startSession();
-  }
-
-  @Override public void onActivityPaused(Activity activity) {
-    super.onActivityPaused(activity);
-    amplitude.endSession();
   }
 
   @Override public void identify(IdentifyPayload identify) {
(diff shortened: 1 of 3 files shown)


analytics-integrations/amplitude/src/main/java/com/segment/analytics/internal/integrations/AmplitudeIntegration.java
```diff
@@ -48,12 +53,10 @@
 
   @Override public void onActivityResumed(Activity activity) {
```
f2prateek: This method is now empty so it can be deleted.

analytics-integrations/amplitude/src/main/java/com/segment/analytics/internal/integrations/AmplitudeIntegration.java
```diff
@@ -48,12 +53,10 @@
...
     super.onActivityResumed(activity);
-    amplitude.startSession();
   }
 
   @Override public void onActivityPaused(Activity activity) {
```
f2prateek: This method is now empty so it can be deleted.

analytics-integrations/amplitude/src/test/java/com/segment/analytics/internal/integrations/AmplitudeTest.java
```diff
@@ -112,16 +112,12 @@
     Activity activity = mock(Activity.class);
 
     integration.onActivityResumed(activity);
-
-    verify(amplitude).startSession();
```
f2prateek: Verify that there were no interactions with the Amplitude SDK.

analytics-integrations/amplitude/src/test/java/com/segment/analytics/internal/integrations/AmplitudeTest.java
```diff
@@ -112,16 +112,12 @@
...
   }
 
   @Test public void activityPause() {
     Activity activity = mock(Activity.class);
 
     integration.onActivityPaused(activity);
-
-    verify(amplitude).endSession();
```
f2prateek: Verify that there were no interactions with the Amplitude SDK.

f2prateek:
LGTM, few inline changes.

We'd also need some more info about the new settings so we can share it in our web UI.

Something like this:
```
"trackNamedPages": {
      "default": true,
      "description": "This will track events to Amplitude for [`page` method](https://segment.io/libraries/analytics.js#page) calls that have a `name` associated with them. For example `page('Signup')` would translate to **Viewed Signup Page**.",
      "label": "Track Named Pages to Amplitude",
      "type": "boolean"
    },
```

analytics-integrations/amplitude/src/main/java/com/segment/analytics/internal/integrations/AmplitudeIntegration.java
djih: Deleted

analytics-integrations/amplitude/src/main/java/com/segment/analytics/internal/integrations/AmplitudeIntegration.java
djih: Deleted

analytics-integrations/amplitude/src/test/java/com/segment/analytics/internal/integrations/AmplitudeTest.java
djih: Done

analytics-integrations/amplitude/src/test/java/com/segment/analytics/internal/integrations/AmplitudeTest.java
djih: Done

f2prateek:
Perfect, please squash the commits into one.

djih:
Squashed.

"trackSessionEvents": {
      "default": false,
      "description": "This will enable the sending of start and end session events to Amplitude, which are [no longer sent by default](https://github.com/amplitude/Amplitude-Android#tracking-sessions).",
      "label": "Track Session Events to Amplitude",
      "type": "boolean"
    }

PR #342, opened 2015-09-03
Improve GA integration

f2prateek:
* Add support for custom dimensions and metrics.
* Send a regular event in addition to ecommerce events for any ecommerce
  calls.
* Dont set screenName to null.
* Traits/Properties are only set on if they represent special
  values or are mapped by custom dimensions/metrics.

Closes #333


sperand-io:
boom!

PR #349, opened 2015-09-14
Add 'collectDeviceId' Builder option

jtomson-mdsol:
This PR adds an option to the `Builder`, `collectDeviceId`.

If the option is disabled, the `device:id` property of the device json object will use the `anonymousId` in place of the device identifier. This option is enabled by default.

Some terms-of-service do not allow this type of collection to take place, so it helps our adoption of Segment knowing we can disable it.

### DIFF
--- a/analytics-core/src/main/java/com/segment/analytics/Analytics.java
+++ b/analytics-core/src/main/java/com/segment/analytics/Analytics.java
@@ -579,6 +579,7 @@ public class Analytics {
 
     private final Application application;
     private String writeKey;
+    private boolean collectDeviceID = Utils.DEFAULT_COLLECT_DEVICE_ID;
     private int flushQueueSize = Utils.DEFAULT_FLUSH_QUEUE_SIZE;
     private long flushIntervalInMillis = Utils.DEFAULT_FLUSH_INTERVAL;
     private Options defaultOptions;
@@ -640,6 +641,18 @@ public class Analytics {
         throw new IllegalArgumentException("flushInterval must be greater than zero.");
       }
       this.flushIntervalInMillis = timeUnit.toMillis(flushInterval);
+      return this;
+    }
+
+    /**
+     * Enable or disable collection of {@link android.provider.Settings.Secure#ANDROID_ID},
+     * {@link android.os.Build#SERIAL} or the Telephony Identifier retreived via
+     * {@link TelephonyManager#getDeviceId()} as available.
+     *
+     * Collection of the device identifier is enabled by default.
+     */
+    public Builder collectDeviceId(boolean collect) {
+      this.collectDeviceID = collect;
       return this;
     }
 
@@ -759,7 +772,8 @@ public class Analytics {
         Traits traits = Traits.create();
         traitsCache.set(traits);
       }
-      AnalyticsContext analyticsContext = AnalyticsContext.create(application, traitsCache.get());
+      AnalyticsContext analyticsContext = AnalyticsContext.create(application, traitsCache.get(),
+              this.collectDeviceID);
       analyticsContext.attachAdvertisingId(application);
 
       synchronized (INSTANCES) {
--- a/analytics-core/src/main/java/com/segment/analytics/Analytics.java
+++ b/analytics-core/src/main/java/com/segment/analytics/Analytics.java
@@ -773,7 +773,7 @@ public class Analytics {
         traitsCache.set(traits);
       }
       AnalyticsContext analyticsContext = AnalyticsContext.create(application, traitsCache.get(),
-              this.collectDeviceID);
+              collectDeviceID);
       analyticsContext.attachAdvertisingId(application);
 
       synchronized (INSTANCES) {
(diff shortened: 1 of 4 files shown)


analytics-core/src/main/java/com/segment/analytics/AnalyticsContext.java
```diff
@@ -214,9 +226,10 @@ public Campaign campaign() {
   }
 
   /** Fill this instance with device info from the provided {@link Context}. */
-  void putDevice(Context context) {
+  void putDevice(Context context, boolean collectDeviceID) {
     Device device = new Device();
-    device.put(Device.DEVICE_ID_KEY, getDeviceId(context));
+    String identifier = collectDeviceID ? getDeviceId(context) : traits().anonymousId();
```
f2prateek: I think we should generate a random id instead of using the anonymousId.

analytics-core/src/main/java/com/segment/analytics/AnalyticsContext.java
```diff
@@ -115,23 +116,34 @@
...
     return analyticsContext;
   }
 
+  /**
+   * Create a new {@link AnalyticsContext} instance filled in with information from the given
+   * {@link Context}. The {@link Analytics} client can be called from anywhere, so the returned
+   * instances is thread safe. This is a convenience wrapper for the method
+   * {@link #create(Context, Traits, boolean)}
+   */
+  static synchronized AnalyticsContext create(Context context, Traits traits) {
```
f2prateek: We don't need both the methods defined. Let's just use the one that keeps the `boolean` variant.

analytics-core/src/main/java/com/segment/analytics/AnalyticsContext.java
```diff
@@ -115,23 +116,34 @@
...
     analyticsContext.putLibrary();
     analyticsContext.put(LOCALE_KEY,
-        Locale.getDefault().getLanguage() + "-" + Locale.getDefault().getCountry());
+            Locale.getDefault().getLanguage() + "-" + Locale.getDefault().getCountry());
     analyticsContext.putNetwork(context);
     analyticsContext.putOs();
     analyticsContext.putScreen(context);
     putUndefinedIfNull(analyticsContext, USER_AGENT_KEY, System.getProperty("http.agent"));
     putUndefinedIfNull(analyticsContext, TIMEZONE_KEY, TimeZone.getDefault().getID());
-    analyticsContext.setTraits(traits);
```
f2prateek: This should be left in.

f2prateek: ah nvm, I see you moved it above.

analytics-core/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -759,7 +772,8 @@ public Analytics build() {
         Traits traits = Traits.create();
         traitsCache.set(traits);
       }
-      AnalyticsContext analyticsContext = AnalyticsContext.create(application, traitsCache.get());
+      AnalyticsContext analyticsContext = AnalyticsContext.create(application, traitsCache.get(),
+              this.collectDeviceID);
```
f2prateek: `this.` is unnecessary here.

analytics-core/src/main/java/com/segment/analytics/AnalyticsContext.java
jtomson-mdsol: That's the route I went initially, but then I worried that some integrations may treat that as separate devices for every session?

analytics-core/src/main/java/com/segment/analytics/AnalyticsContext.java
jtomson-mdsol: sounds good!

analytics-core/src/main/java/com/segment/analytics/Analytics.java
jtomson-mdsol: :+1:

analytics-core/src/main/java/com/segment/analytics/AnalyticsContext.java
f2prateek: We could cache it, or alternatively, just leave it empty rather than make up something?

f2prateek: On second thought, I think what you have will work, I'm just overthinking it.

jtomson-mdsol:
sorry not sure why the circleci checks are failing - the tests seem to be passing for me locally

analytics-core/src/main/java/com/segment/analytics/AnalyticsContext.java
```diff
@@ -115,20 +116,21 @@
...
     AnalyticsContext analyticsContext =
         new AnalyticsContext(new NullableConcurrentHashMap<String, Object>());
     analyticsContext.putApp(context);
-    analyticsContext.putDevice(context);
+    analyticsContext.setTraits(traits);
+    analyticsContext.putDevice(context, collectDeviceId);
     analyticsContext.putLibrary();
     analyticsContext.put(LOCALE_KEY,
-        Locale.getDefault().getLanguage() + "-" + Locale.getDefault().getCountry());
+            Locale.getDefault().getLanguage() + "-" + Locale.getDefault().getCountry());
```
f2prateek: The CI is failing with a checkstyle error — likely because of this line. You can run `./gradlew build` locally to run checkstyle and verify the cause.

jtomson-mdsol: whoops - thanks should be fixed

jtomson-mdsol:
> sorry not sure why the circleci checks are failing 
should be good to go now

PR #350, opened 2015-09-16
ensure only US ISO 8601 timestamps

graingert:


graingert:
> The main reason you'd create an instance this class directly is because you need to format/parse a specific machine-readable format, in which case you almost certainly want to explicitly ask for US to ensure that you get ASCII digits (rather than, say, Arabic digits). (See "Be wary of the default locale".) The most useful non-localized pattern is "yyyy-MM-dd HH:mm:ss.SSSZ", which corresponds to the ISO 8601 international standard date format. 

https://developer.android.com/reference/java/text/SimpleDateFormat.html

f2prateek:
Thanks, what does the zzzz do?

graingert:
Complies with ISO 8601 timezone offsets

f2prateek:
I don't think we need the extended `ZZZZZ` over the basic `Z`, so let's leave that part in as the same for now.

graingert:
I'll make a separate PR for that

graingert:
@f2prateek removed that commit.

graingert:
@f2prateek We only use the Android and iOS libraries, so it might be a good idea to double check your other libraries for this problem.

graingert:
@f2prateek do you know when we might get a release that includes this change?

f2prateek:
@graingert I'll try to release one today.

graingert:
@f2prateek can we get a release please?

PR #352, opened 2015-09-16
ensure only US ISO 8601 timestamps

graingert:


graingert:
> 
    z - [short form] specific non-location
        falling back to localized GMT
    zzzz - [long form] specific non-location
        falling back to localized GMT
    Z - RFC 822 (no fallback necessary)
    ZZZZ - Localized GMT (no fallback necessary)
    ZZZZZ - ISO 8601 (no fallback necessary)
    V - specific non-location
        falling back to localized GMT
From http://www.unicode.org/reports/tr35/tr35-25.html#Time_Zone_Fallback

f2prateek:
I think the extended version is unnecessary, plus our servers will automatically revert to the basic one when they parse the timestamps, so this won't make a difference.

PR #367, opened 2015-10-04
updating Amplitude SDK to v2.1.0

djih:
A bug fix and added support for user property operations: https://github.com/amplitude/Amplitude-Android/releases/tag/v2.1.0.

The underlying implementation for setUserProperties now uses our identify api, but the Segment Identify call should still work as intended.

### DIFF
--- a/analytics-integrations/amplitude/build.gradle
+++ b/analytics-integrations/amplitude/build.gradle
@@ -1,5 +1,5 @@
 apply from: rootProject.file('gradle/integration.gradle')
 
 dependencies {
-  compile 'com.amplitude:android-sdk:2.0.4'
+  compile 'com.amplitude:android-sdk:2.1.0'
 }


f2prateek:
@djih should we also switch our integration to directly use the identify API? Are there any advantages/disadvantages?

djih:
@f2prateek: our identify api is simply just [allowing these user property operations](https://github.com/amplitude/Amplitude-Android/tree/v2.1.0#user-property-operations). The idea is any changes to user properties are declared via our Identify object. You can set individual properties, or add, unset, setOnce, etc.

Segment's Identify is setting a bunch of user properties at once. It's fine to just call our setUserProperties, as it will iterate through all the key:value pairs, and [set them on a new Identify object](https://github.com/amplitude/Amplitude-Android/blob/v2.1.0/src/com/amplitude/api/AmplitudeClient.java#L628-L638). It's perfectly fine for you guys to keep using setUserProperties. If you were to use our identify API directly, then you would have to do something similar (iterate through all the key/values, and set them on an identify object).

It doesn't seem like your API takes advantage of operations like setOnce, unset, and add. That's the only downside I guess.

djih:
@f2prateek : forgot to mention, my above comment applies to our JS integration as well (I just created a PR for that update as well). I am still working on the iOS version of identify

PR #388, opened 2015-11-05
Changed Localytics LogLevel Trigger

chrisjenx:
Localytics internal logging outputs at verbose level. It makes sense for us to do the same from Segment to trigger that. Seen as Localytics is really verbose this probably favours most implementations too.

### DIFF
--- a/analytics-integrations/localytics/src/main/java/com/segment/analytics/internal/integrations/LocalyticsIntegration.java
+++ b/analytics-integrations/localytics/src/main/java/com/segment/analytics/internal/integrations/LocalyticsIntegration.java
@@ -42,7 +42,7 @@ public class LocalyticsIntegration extends AbstractIntegration<Void> {
       throws IllegalStateException {
     log = analytics.getLogger().newLogger(LOCALYTICS_KEY);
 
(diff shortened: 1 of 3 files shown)


f2prateek:
Thanks! We've made some changes to the structure of the repos and will be migrating the integrations to live in their own repos, I'll make the update in the localytics repo manually.

f2prateek:
Added to https://github.com/segment-integrations/analytics-android-integration-localytics

chrisjenx:
Cool thanks Prateek.

What was the reason to separate out there packages?

f2prateek:
It was starting to become unmaintainable to have all the integrations live under one repo (we had the same problem with analytics.js, which used to contain 50+ integrations in it's repo). A significant majority of the updates we've made to analytics-android in the last year have been integration fixes, and splitting these out will avoid unnecessary churn in the core library itself.

Additionally we want partners to be able to own their integrations. Currently, Segment has to merge each PR submitted by a partner and make a release to the entire library. The split allows partners to host the integration in their repos, and gives them to ability make changes and releases independent of us.

This similar to how Retrofit don't provide every possible converter inside the core library as a module, but rather encourages converters to be provided as their own [library](https://github.com/aurae/retrofit-logansquare).

chrisjenx:
OK cool. Good to know. :)

PR #411, opened 2016-02-25
Flush events on interval over wifi or wait until queue size met

astapley:


graingert:
Android 6 decouples "is unlimited data" from "is WiFi".

This PR does not take into account when devices are connected to a GSM WiFi hotspot that is pay per GiB.

f2prateek:
Yeah I don't think we want to add too much magic here. We try to have reasonable defaults, and you can always have your own logic in say, a service, that calls `analytics.flush()` when you detect you're on WiFi for example.

astapley:
@graingert anything you can link me to that explains this?

astapley:
@f2prateek why would this add too much magic? I think the amount of data being transferred and trying to optimally send it over wifi is something to consider.Yes, I could manually flush over wifi myself but that seems a hacky solution to something Segment should handle. If you have any other suggestions for how it should be handled by Segment, I'd be happy to work in that direction.

f2prateek:
I'm not sure why you think Segment should handle this. Every app has different requirements. The main reason we batch is to preserve battery life and reduce the network requests we make. This change will adversely affect both.

PR #483, opened 2016-11-14
Revert relying on HTTP Cache for cache settings introduced in 4.2.2 to avoid interfering  with application level cache behavior.

tonyxiao:


f2prateek:
Sorry missed this. Closing in favor of https://github.com/segmentio/analytics-android/pull/484 which is now merged.

PR #486, opened 2016-12-20
Remove unnecessary bounded wildcard

NightlyNexus:


f2prateek:
`? extends String` makes no sense, but unfortunately the implementing the Map interface requires it  http://docs.oracle.com/javase/8/docs/api/java/util/Map.html#putAll-java.util.Map-

From https://circleci.com/gh/segmentio/analytics-android/804
```
 /home/ubuntu/analytics-android/analytics/src/main/java/com/segment/analytics/ValueMap.java:32: error: ValueMap is not abstract and does not override abstract method putAll(Map<? extends String,? extends Object>) in Map
public class ValueMap implements Map<String, Object> {
       ^
/home/ubuntu/analytics-android/analytics/src/main/java/com/segment/analytics/ValueMap.java:98: error: name clash: putAll(Map<String,?>) in ValueMap and putAll(Map<? extends K,? extends V>) in Map have the same erasure, yet neither overrides the other
  @Override public void putAll(Map<String, ?> map) {
                        ^
  where K,V are type-variables:
    K extends Object declared in interface Map
    V extends Object declared in interface Map
/home/ubuntu/analytics-android/analytics/src/main/java/com/segment/analytics/ValueMap.java:98: error: method does not override or implement a method from a supertype
  @Override public void putAll(Map<String, ?> map) {
  ^
```

NightlyNexus:
i don't know what i was doing. yep, haha, 😶

PR #523, opened 2017-04-11
Unregister ActivityLifecycleCallbacks on Analytics shutdown

stagg:
Resolves an issue where the registered ActivityLifecycleCallbacks could be triggered after Analytics has been shutdown. This would result in a RejectedExecutionException when a IntegrationOperation was submitted to a shutdown _analyticsExecutor_.


f2prateek:
@stagg thanks! could we add a test for this?

stagg:
Sure thing!

PR #524, opened 2017-04-18
Unregister ActivityLifecycleCallbacks on shutdown.

f2prateek:
This ensures that no lifecycle callbacks are queued on a executor that
has been shutdown.

Closes #523

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/Analytics.java
+++ b/analytics/src/main/java/com/segment/analytics/Analytics.java
@@ -126,6 +126,7 @@ public class Analytics {
   final Cartographer cartographer;
   private final ProjectSettings.Cache projectSettingsCache;
   final Crypto crypto;
+  @Private final Application.ActivityLifecycleCallbacks activityLifecycleCallback;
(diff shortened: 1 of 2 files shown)


analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -438,6 +440,9 @@ void recordScreenViews(Activity activity) {
 
   @Private
   void runOnMainThread(final IntegrationOperation operation) {
+    if (shutdown) {
```
f2prateek: @stagg was wondering why this was added. This shouldn't be invoked if the shutdown code does it's job right?

stagg: It's not necessary, and one of the tests uses this to verify the activity callbacks don't propagate. I was just being extra defensive as runOnMainThread _could_ be called from a few other places.

f2prateek: Fair enough!

PR #529, opened 2017-06-01
Project cleanups

friederbluemle:
A bunch of project cleanups.

<sub><sup>Please NO GitHub rebase/squash merges. If you'd like me to combine some or all of the commits, please let me know, and I will update my branch. Thank you.</sup></sub>


.gitignore
```diff
@@ -1,151 +1,24 @@
-# Created by http://www.gitignore.io
```
f2prateek: I'd prefer not to remove things from this list just yet. In terms of additions, some seem unnecessary for this project. If they were autogenerated by some tool like gitignore.io that would be ok, but if they're by hand I'd prefer to remove them if they're unused.

f2prateek:
I think the interesting bits are:

1. Update Android Gradle plugin to 2.3.2	
2. Update Gradle wrapper to 3.5	
3. Update build tools to 25.0.3
4. Tweak jvmargs.

These all look independent, so I'd prefer to see them as seperate PRs. I'm not too attached to 4) as well, it gets rid of a warning, but the number seems too magical..

.gitignore
friederbluemle: Good suggestions! The new .gitignore file is a simplified version of the "official" file from github/gitignore. - The main issue with some predefined gitignore files is that they declare TONS of unnecessary and redundant stuff. Like 90% of the lines were never used. You're right that some of the additions are in fact unnecessary, for example the google-services.json file (your project does not use Firebase). I have removed it and amended the commit.

friederbluemle:
@f2prateek Thanks for the review. Yes, absolutely! I have moved the dependency update commits to a separate branch, and will open a separate PR from the rebased new branch immediately after this PR is merged. I also updated the title of this PR, since it's now just cleanups (no dependency updates). Further, I added a new commit to remove the executable bits from regular files (I am not sure why they were set).

f2prateek:
Sorry but we don't really need most of these. We could use the ones changing the executable bits (I think the `./gradlew bat` needs to still be executable).

friederbluemle:
@f2prateek No worries. I saw you redid most of the changes yourself. If you don't like contributions like this, maybe just make it clear in a `CONTRIBUTING.md` file or similar. It would save me (and potentially others) some time. Thanks :)

f2prateek:
Sorry about that. Hope tit doesn't dissuade you from submitting future prs. Generally happy to accept changes like these if they're small in scope. Big PRs that tackle multiple problems are generally a recipe for disaster.

PR #536, opened 2017-07-18
Collect Advertising ID from Amazon Fire devices.

f2prateek:
Reference:
https://developer.amazon.com/public/solutions/devices/fire-tv/docs/fire-tv-advertising-id#verifying-and-responding-to-the-users-advertising-id-choices


f2prateek:
cc @ladanazita @tonyxiao

PR #537, opened 2017-07-23
Fix calling to getInputStream on errors

yoavst:
When the status code is above 300 (Don't know for sure if for all codes, but for some of them for sure), `getInputStream()` throws `FileNotFoundException` and the stream is in `getErrorStream()`.


f2prateek:
Is this documented anywhere?

yoavst:
https://stackoverflow.com/questions/4633048/httpurlconnection-reading-response-content-on-403-error

PR #547, opened 2017-10-24
Rename getter "putSubtotal()" to "subtotal()"

egor-n:


f2prateek:
Thanks! Could you leave the old method in and mark it as deprecated? Don't want to accidentally break API for folks that might be using this - we can remove the deprecated one in a couple of months

f2prateek:
Thanks!

PR #551, opened 2017-10-30
Add jcenter to allprojects/repositories block in top-level build.gradle

brennan:
Adding jcenter() to the top-level build.gradle file will help eliminate confusion on the Success Team - I just came from a meeting where they weren't able to compile dependencies into our sample projects because the jcenter() repo was not included anywhere. This PR solves that problem.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,7 @@ allprojects {
 
   repositories {
     mavenCentral()
+    jcenter()
   }
 
   group = GROUP


f2prateek:
Why is jcenter required here?

I don't think jcenter is not required for this project. I believe it's only required for certain integrations. I don't think it's a good idea to add integration specific requirements here.

I also think adding jcenter will only increase confusion. Customers still have to add it by hand for if their tools require it. I think this will lead to confusion from a customer who didn't add jcenter saying it doesn't work - but our success team is using the sample app from here and says it does work.

If a tool requires jcenter, we should document it in the integration setup for that tool.

brennan:
Ok - fair enough!

PR #552, opened 2017-11-06
Support schema defaults for new events.

f2prateek:
If an event is not present in the tracking plan, we'll now lookup the schema defaults for the project.

If the event is disabled in the schema default, it will only be sent to the Segment integration (i.e. api.segment.io).

If the event is enabled in the schema default, it will be sent to all integrations.

Ref: https://paper.dropbox.com/doc/Schema-Client-Side-Defaults-DufdS8Ej43mnFXvvMqm1b


f2prateek:
Made some test updates to get better coverage.

PR #573, opened 2018-02-03
Set the userId on the analytics executor thread.

### LINKED ISSUE #522: UserID should be recorded at time of event.
While using Segment with Mixpanel integration for Android, I faced a racing condition when calling `alias` followed by `identify` like in the documentation described here: [Alias](https://segment.com/docs/sources/mobile/android/#alias)

Expected outcome: 
`anonymousId` is aliased to `userId` and then identified with `userId`

`analytics.alias("anonymousId", "userId");`
`analytics.identify("userId");`

But outcome is sometimes: 
`analytics.alias("userId", "userId");`
`analytics.identify("userId");`

So it seems like that even when `identify` is called after `alias`,  `alias` picks up the `userId` set by `identify` instead of `anonymousId`

f2prateek:
Previously we were setting the userId on the thread that `identify` is called (which may be any thread), but reading the userId on a background thread.

Because of this, the userId on events called before identify may have the userId from "newer" events.

```
analytics.identify("foo");
analytics.track("event"); -> may have userId foo
analytics.identify("bar");
```

To fix this, we now also set the userId on the background thread.

Originally referenced in https://github.com/segmentio/analytics-android/issues/522. Note that my original solution proposed we read the userId on the main thread to fix the issue, but the fix implemented here is to read the userId on the background thread instead.

Ref: LIB-120

Closes #522

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/Analytics.java
+++ b/analytics/src/main/java/com/segment/analytics/Analytics.java
@@ -491,27 +491,29 @@ public class Analytics {
    * @see <a href="https://segment.com/docs/spec/identify/">Identify Documentation</a>
    */
   public void identify(
-      @Nullable String userId, @Nullable Traits newTraits, final @Nullable Options options) {
+      final @Nullable String userId,
+      final @Nullable Traits newTraits,
+      final @Nullable Options options) {
     assertNotShutdown();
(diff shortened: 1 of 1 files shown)


flagbug:
@f2prateek Is there a plan to ever release this fix in a stable version? I think we're hitting this issue, but since we've pinned the Segment library to stable versions, it looks like this has never been picked up in our app...

PR #612, opened 2019-04-10
Add settings to middlewares

CarlosMecha:
This change adds a new method to the `Chain` interface to retrieve the project settings in a middleware interceptor.

My use case is, we are creating a Adobe Marketing Cloud Visitior ID middleware and we need to make a request to an external HTTP API to retrieve and then inject the ID to Segment events. For that request, we need the value of the Organization ID (a setting already present and rendered on CDN).

Please let me know if this makes sense. If not, the customer using the middleware needs to provide the Organization ID when instantiating the middleware.

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/Analytics.java
+++ b/analytics/src/main/java/com/segment/analytics/Analytics.java
@@ -1298,9 +1298,7 @@ public class Analytics {
       return this;
     }
 
-    /**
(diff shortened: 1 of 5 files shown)


CarlosMecha:
Talked online with Prateek. Since this is a breaking change (on `Chain` interface), we'll do something else.

PR #620, opened 2019-07-16
Add Application Open and Application Backgrounded Events

carlos-kelly:
Addresses LIB-1156 and LIB-1143.

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/Analytics.java
+++ b/analytics/src/main/java/com/segment/analytics/Analytics.java
@@ -76,6 +76,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * The entry point into the Segment for Android SDK.
@@ -291,11 +292,18 @@ public class Analytics {
     activityLifecycleCallback =
         new Application.ActivityLifecycleCallbacks() {
           final AtomicBoolean trackedApplicationLifecycleEvents = new AtomicBoolean(false);
+          final AtomicInteger numberOfActivities = new AtomicInteger(1);
+          final AtomicBoolean isChangingActivityConfigurations = new AtomicBoolean(false);
+          final AtomicBoolean firstLaunch = new AtomicBoolean(false);
 
           @Override
           public void onActivityCreated(Activity activity, Bundle savedInstanceState) {
+            runOnMainThread(IntegrationOperation.onActivityCreated(activity, savedInstanceState));
+
             if (!trackedApplicationLifecycleEvents.getAndSet(true)
                 && shouldTrackApplicationLifecycleEvents) {
+              numberOfActivities.set(0);
+              firstLaunch.set(true);
               trackApplicationLifecycleEvents();
 
               if (trackAttributionInformation) {
@@ -308,7 +316,6 @@ public class Analytics {
                     });
               }
             }
-            runOnMainThread(IntegrationOperation.onActivityCreated(activity, savedInstanceState));
           }
 
           @Override
@@ -322,32 +329,36 @@ public class Analytics {
           @Override
           public void onActivityResumed(Activity activity) {
             runOnMainThread(IntegrationOperation.onActivityResumed(activity));
-            if (shutdown) {
-              return;
+
+            if (shouldTrackApplicationLifecycleEvents
+                && numberOfActivities.incrementAndGet() == 1
+                && !isChangingActivityConfigurations.get()) {
+
+              Properties properties = new Properties();
+              if (firstLaunch.get()) {
+                properties.putValue("version", currentVersion).putValue("build", currentBuild);
+              }
+              properties.putValue("from_background", !firstLaunch.getAndSet(false));
+
+              track("Application Opened", properties);
             }
-            track(
-                "Application Opened",
-                new Properties()
-                    .putValue("version", currentVersion)
-                    .putValue("build", currentBuild));
           }
 
           @Override
           public void onActivityPaused(Activity activity) {
             runOnMainThread(IntegrationOperation.onActivityPaused(activity));
-            if (shutdown) {
-              return;
-            }
-            track(
-                "Application Backgrounded",
-                new Properties()
-                    .putValue("version", currentVersion)
-                    .putValue("build", currentBuild));
           }
 
           @Override
           public void onActivityStopped(Activity activity) {
             runOnMainThread(IntegrationOperation.onActivityStopped(activity));
+
+            isChangingActivityConfigurations.set(activity.isChangingConfigurations());
+            if (shouldTrackApplicationLifecycleEvents
+                && numberOfActivities.decrementAndGet() == 0
+                && !isChangingActivityConfigurations.get()) {
+              track("Application Backgrounded");
+            }
           }
 
           @Override
@@ -424,13 +435,6 @@ public class Analytics {
               .putValue("previous_" + VERSION_KEY, previousVersion)
               .putValue("previous_" + BUILD_KEY, previousBuild));
     }
-
-    // Track Application Opened.
-    track(
-        "Application Opened",
-        new Properties() //
-            .putValue("version", currentVersion) //
-            .putValue("build", currentBuild));
 
     // Update the recorded version.
     SharedPreferences.Editor editor = sharedPreferences.edit();
--- a/analytics/src/main/java/com/segment/analytics/Analytics.java
+++ b/analytics/src/main/java/com/segment/analytics/Analytics.java
@@ -288,6 +288,10 @@ public class Analytics {
         new Application.ActivityLifecycleCallbacks() {
           final AtomicBoolean trackedApplicationLifecycleEvents = new AtomicBoolean(false);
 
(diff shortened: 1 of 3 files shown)


analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -318,11 +322,27 @@ public void onActivityStarted(Activity activity) {
...
+                    .putValue("build", currentBuild));
           }
 
           @Override
           public void onActivityPaused(Activity activity) {
             runOnMainThread(IntegrationOperation.onActivityPaused(activity));
+            if (shutdown) {
+              return;
+            }
+            track(
```
f2prateek: Same as https://github.com/segmentio/analytics-ios/pull/831#discussion_r304657458 - I don't believe any properties should be collected here, as per https://segment.com/docs/spec/mobile/#application-backgrounded.

carlos-kelly:
That is a good point. Since we don't have access to `android.arch.lifecycle` I could use the atomic boolean like we did for `onActivityCreated` to ensure it only fires for the first activity. Otherwise the user would have to manually add the tracking.

f2prateek:
👍 I think there are tradeoffs of both approaches:

* `android.arch.lifecycle` approach: likely more reliable as they've handled edge cases for us, but requires adding a new dependency.
* boolean approach: likely less reliable as we might not account properly for all edge cases, but doesn't require a dependency.

I think the boolean approach is best for now as it mimics what we already do, but we should consider the `android.arch.lifecycle` approach in the future to make this more bullet proof.

carlos-kelly:
Sounds good. I will update that!

carlos-kelly:
It looks like it is part of Jetpack AndroidX so there are other dependencies to add when using `android.arch.lifecycle`. https://developer.android.com/jetpack/androidx/releases/lifecycle

f2prateek:
Oof that sucks - yeah I think the boolean approach is definitely the right one for now!

analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -284,14 +285,25 @@ public void run() {
...
+    PackageInfo packageInfo = getPackageInfo(application);
+    final String currentVersion = packageInfo.versionName;
+    final int currentBuild = packageInfo.versionCode;
+
     activityLifecycleCallback =
         new Application.ActivityLifecycleCallbacks() {
           final AtomicBoolean trackedApplicationLifecycleEvents = new AtomicBoolean(false);
+          final AtomicInteger numberOfActivities = new AtomicInteger(1);
+          final AtomicBoolean isChangingActivityConfigurations = new AtomicBoolean(false);
+          final AtomicBoolean firstLaunch = new AtomicBoolean(false);
```
carlos-kelly: Android has new application lifecycle callbacks via `androidx.lifecycle`, but we can't use them due to the dependency requirements AndroidX places. Additionally that breaks the React Native wrapper as AndroidX deps won't work with new RN projects out of the box. We use the traditional method of activity reference counting to detect if we are launching a new activity or resuming to foreground.

analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -284,14 +285,25 @@ public void run() {
...
 
           @Override
           public void onActivityCreated(Activity activity, Bundle savedInstanceState) {
+            runOnMainThread(IntegrationOperation.onActivityCreated(activity, savedInstanceState));
+
             if (!trackedApplicationLifecycleEvents.getAndSet(true)
                 && shouldTrackApplicationLifecycleEvents) {
+              numberOfActivities.set(0);
+              firstLaunch.set(true);
```
carlos-kelly: These reset the atomic variables for native Android apps. `onActivityCreated` is not called when this library is used by the React Native wrapper.

analytics/src/test/java/com/segment/analytics/AnalyticsTest.java
```diff
@@ -896,19 +883,6 @@ protected boolean matchesSafely(TrackPayload payload) {
                         payload.properties().getInt("build", -1) == 101;
                   }
                 }));
-    verify(integration)
```
f2prateek: why was this removed?

analytics/src/test/java/com/segment/analytics/AnalyticsTest.java
```diff
@@ -796,19 +796,6 @@ protected boolean matchesSafely(TrackPayload payload) {
                         payload.properties().getInt("build", -1) == 100;
                   }
                 }));
-    verify(integration)
```
f2prateek: why was this removed?

analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -318,6 +329,19 @@ public void onActivityStarted(Activity activity) {
...
+
+            if (shouldTrackApplicationLifecycleEvents
+                && numberOfActivities.incrementAndGet() == 1
+                && !isChangingActivityConfigurations.get()) {
+
+              Properties properties = new Properties();
+              if (firstLaunch.get()) {
+                properties.putValue("version", currentVersion).putValue("build", currentBuild);
+              }
+              properties.putValue("from_background", !firstLaunch.getAndSet(false));
```
carlos-kelly: On first launch we want to ensure `from_background` is false, subsequent calls, this will be true.

analytics/src/test/java/com/segment/analytics/AnalyticsTest.java
carlos-kelly: Both were removed as `ApplicationOpened` is no longer called during `trackApplicationLifecycleEventsInstalled`. It is now called in `onActivityResumed` and depending if it's a cold start or not, the correct Properties are passed.

analytics/src/test/java/com/segment/analytics/AnalyticsTest.java
```diff
@@ -1050,6 +1023,205 @@ protected boolean matchesSafely(Application.ActivityLifecycleCallbacks item) {
...
+    verify(integration)
+        .track(
+            argThat(
+                new NoDescriptionMatcher<TrackPayload>() {
+                  @Override
+                  protected boolean matchesSafely(TrackPayload payload) {
+                    return payload.event().equals("Application Opened")
+                        && payload.properties().getString("version").equals("1.0.0")
+                        && payload.properties().getInt("build", -1) == 100
+                        && !payload.properties().getBoolean("from_background", true);
```
carlos-kelly: This is the first integration test that ensures that on first start `from_background` is false, the correct build params are part of the cold start `Application Opened` tracking event.

analytics/src/test/java/com/segment/analytics/AnalyticsTest.java
```diff
@@ -1050,6 +1023,205 @@ protected boolean matchesSafely(Application.ActivityLifecycleCallbacks item) {
...
+    callback.get().onActivityResumed(null);
+    callback.get().onActivityStopped(backgroundedActivity);
+
+    verify(integration)
+        .track(
+            argThat(
+                new NoDescriptionMatcher<TrackPayload>() {
+                  @Override
+                  protected boolean matchesSafely(TrackPayload payload) {
+                    return payload.event().equals("Application Backgrounded");
```
carlos-kelly: This is the second integration test that launches the app then sends it to background.

analytics/src/test/java/com/segment/analytics/AnalyticsTest.java
```diff
@@ -1050,6 +1023,205 @@ protected boolean matchesSafely(Application.ActivityLifecycleCallbacks item) {
...
+                }));
+
+    verify(integration)
+        .track(
+            argThat(
+                new NoDescriptionMatcher<TrackPayload>() {
+                  @Override
+                  protected boolean matchesSafely(TrackPayload payload) {
+                    return payload.event().equals("Application Opened")
+                        && payload.properties().getBoolean("from_background", false);
```
carlos-kelly: This is the third integration test that verifies when we send the app to the background, then resume it to foreground the `from_background` param is `true`.

analytics/src/main/java/com/segment/analytics/AnalyticsActivityLifecycleCallbacks.java
```diff
@@ -0,0 +1,163 @@
...
+import android.app.Activity;
+import android.app.Application;
+import android.content.pm.PackageInfo;
+import android.os.Bundle;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class AnalyticsActivityLifecycleCallbacks implements Application.ActivityLifecycleCallbacks {
```
f2prateek: This should be a private class.

f2prateek: (or at least package protected)

analytics/src/test/java/com/segment/analytics/AnalyticsTest.java
f2prateek: Got it - shouldn't this test be moved somewhere else then?

f2prateek: I see it is moved, never mind!

fathyb:
I merged by accident, sorry! Made a revert PR under #621

Issue #616, opened 2019-05-24
SegmentIntegration.flushScheduler not shutdown in Analytics.shutdown


charlesdurham:
Version: 3.4.1

When we switch from our prod to dev environments we switch analytic instances (different API keys) and call shutdown on the old one. 

Doing this causes:

```
    java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.FutureTask@58269de rejected from com.segment.analytics.internal.Utils$AnalyticsNetworkExecutorService@47ce7bf[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 0]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2085)
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:848)
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1394)
        at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:118)
        at com.segment.analytics.SegmentIntegration.submitFlush(SegmentIntegration.java:314)
        at com.segment.analytics.SegmentIntegration$SegmentDispatcherHandler.handleMessage(SegmentIntegration.java:509)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:193)
        at android.os.HandlerThread.run(HandlerThread.java:65)
```

From what I can tell the error is caused because `analytics.shutdown` shuts down the network executor used to flush in `SegmentIntegration` but not the `flushScheduler` which schedules the work periodically. 

This results in trying to perform a flush after the `networkExecutor` is shutdown.

f2prateek:
We haven't seen this issue before, but I've added this to our [internal JIRA](https://segment.atlassian.net/browse/LIB-1168), I would suggest reaching out to our support team so they can keep you updated with the progress.

jlmcdonnell:
If #628 was meant to fix this, I'm still seeing this crash in `4.5.0-beta.0`

prayansh:
@charlesdurham @jlmcdonnell fix should have been released as of `4.5.0-beta.1`.
Gonna close this for now. Reopen if issue still persists

PR #628, opened 2019-08-20
Fixed error where post-shutdown, a flush would be executed and crash …

bsneed:
- Fixes an issue where the segment integration ends up calling submitFlush() after the network executor has been shutdown.  This would cause an exception.
- Added a test to verify that submitFlush() is not called if the network executor is stopped.
- Addresses LIB-621 and #616

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/SegmentIntegration.java
+++ b/analytics/src/main/java/com/segment/analytics/SegmentIntegration.java
@@ -334,15 +334,17 @@ class SegmentIntegration extends Integration<Void> {
       return;
     }
 
-    networkExecutor.submit(
-        new Runnable() {
-          @Override
-          public void run() {
-            synchronized (flushLock) {
-              performFlush();
-            }
-          }
-        });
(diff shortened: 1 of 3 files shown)


analytics/src/main/java/com/segment/analytics/SegmentIntegration.java
```diff
@@ -334,15 +334,17 @@ void submitFlush() {
       return;
     }
 
-    networkExecutor.submit(
-        new Runnable() {
-          @Override
-          public void run() {
-            synchronized (flushLock) {
-              performFlush();
+    if (!networkExecutor.isShutdown()) {
```
f2prateek: to minimize nesting, we can copy the pattern from above:

```
if (networkExecutor.isShutdown()) {
  return;
}

networkExecutor.submit ...
```

Maybe we should emit a log when this happens?

bsneed: ok, will do!

PR #631, opened 2019-09-11
LIB-1233: analytics-android: Improve debug-ability of incorrect conne…

bsneed:
…ctionFactory endpoints in Android.

- Added url validity check in openConnection.
- Throws on invalid URL.
- Throws when project settings are invalid or malformed.
- Added test to verify.
- Logging was investigated and was deemed adequate.


analytics/src/main/java/com/segment/analytics/ConnectionFactory.java
```diff
@@ -79,7 +81,15 @@ public HttpURLConnection attribution(String writeKey) throws IOException {
    * #attribution(String)} and {@link #projectSettings(String)}.
    */
   protected HttpURLConnection openConnection(String url) throws IOException {
-    HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
+    URL requestedURL;
+
+    try {
+      requestedURL = new URL(url);
+    } catch (MalformedURLException e) {
+      throw new IOException("Attempted to use malformed url: " + url);
```
f2prateek: we should propagate the underlying error here.

bsneed: Thanks!  I didn't realize I could include multiple exceptions in the signature.

analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -1394,12 +1398,21 @@ ProjectSettings getSettings() {
...
+    }
     ValueMap integrationSettings = projectSettings.integrations();
+    if (isNullOrEmpty(integrationSettings)) {
+      throw new IOException("There are no integration settings!");
+    }
     integrations = new LinkedHashMap<>(factories.size());
     for (int i = 0; i < factories.size(); i++) {
       Integration.Factory factory = factories.get(i);
       String key = factory.key();
+      if (isNullOrEmpty(key)) {
```
f2prateek: Instead of a runtime check, we should annotate the Factory method to explicitly note that this should return a NonNull value (example https://github.com/segmentio/analytics-android/blob/93c7d5bb09b593440a2347a6089db3e9f74012da/analytics/src/main/java/com/segment/analytics/integrations/GroupPayload.java#L46). This would move the check to compile time instead of runtime.

analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -1394,12 +1398,21 @@ ProjectSettings getSettings() {
     return downloadedSettings;
   }
 
-  void performInitializeIntegrations(ProjectSettings projectSettings) {
+  void performInitializeIntegrations(ProjectSettings projectSettings) throws IOException {
+    if (isNullOrEmpty(projectSettings)) {
+      throw new IOException("ProjectSettings is empty!");
+    }
     ValueMap integrationSettings = projectSettings.integrations();
+    if (isNullOrEmpty(integrationSettings)) {
```
f2prateek: I think it's ok for this to null. If it's null, we can treat this is the same as empty map (which would represent the case when no client side integrations are enabled).

analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -1394,12 +1398,21 @@ ProjectSettings getSettings() {
     return downloadedSettings;
   }
 
-  void performInitializeIntegrations(ProjectSettings projectSettings) {
+  void performInitializeIntegrations(ProjectSettings projectSettings) throws IOException {
+    if (isNullOrEmpty(projectSettings)) {
```
f2prateek: AssertionError might be more appropriate here.

analytics/src/main/java/com/segment/analytics/Analytics.java
bsneed: that appears to be marked @nonnull already.

f2prateek: @bsneed yes, that was an example. Specifically here, we should mark the `Integration.factory.key` method as NonNull (https://github.com/segmentio/analytics-android/blob/master/analytics/src/main/java/com/segment/analytics/integrations/Integration.java#L49)

bsneed: Ah, I misunderstood

PR #632, opened 2019-09-18
Make Crypto abstract methods public

jgavris:
It's not currently possible to plug in a custom `Crypto` implementation for storing analytics events on disk since these abstract methods have default / package visibility.


bsneed:
thanks @jgavris !

PR #637, opened 2019-12-07
[DEST-1532] Track "Deep Link Opened" on activity creation

CarlosMecha:
It sends a track call with the following data as properties:
- `url`: Url from the deep link
- `param`...: Url Query param.

The main goal is to send the campaign information from the URL to Segment and destinations like Google Analytics 360.

[DEST-1532](https://segment.atlassian.net/browse/DEST-1532).


bsneed:
@CarlosMecha need to run `gradlew spotlessApply` on your end and push the changes so tests will pass.

CarlosMecha:
Hey @bsneed , I can't run that command, it can't parse my java version. The error is: `FAILURE: Could not determine java version from 11.0.5`

Issue #633, opened 2019-09-30
Alias: not seeing NewTestingUser


markli:
after sent the newId by alias(NewTestingUser), from the console it is: 

analytics.alias("TestingUser", "TestingUser");

aliasunder:
We're also having this problem.

We are using the `react-native-analytics` package and are having this issue issue only with Android, which uses this package. 

When aliasing a user with their new id, the Segment Debugger shows the previous Id passed in twice.

Example:
Previous Id: 39afb4b1-1225-44b9-89e4-3d124dd035eb
New Id: 89a110e6-f27f-4c10-b95a-0b0dba8a0778 

When calling `analytics.alias('89a110e6-f27f-4c10-b95a-0b0dba8a0778 ')` in our code, the Segment Debugger shows that Segment received the following input:

`analytics.alias("39afb4b1-1225-44b9-89e4-3d124dd035eb", "39afb4b1-1225-44b9-89e4-3d124dd035eb");`

I dug into this further and my logs show that the react-native-analytics package is passing the correct values, however `analytics-android` is where things break down. 

Another issue I noticed is that the payload includes an integration that should not be there, as it was not defined in the options.

` "integrations": {
    "library": {
      "name": "analytics-react-native",
      "version": "1.1.0-beta.1"
    }
  }`

I appreciate any input!

prayansh:
@aliasunder and @markli #666 should fix this for you. Thanks for reporting

PR #639, opened 2020-01-10
Fix aliasing by ensuring userId in payload is not overridden by cached userId

aliasunder:
This fixes issue #633 

Currently, if a userId is cached on a device (and therefore stored in `contextCopy.traits().userId()`), this breaks aliasing.

Since `builder.userId` is always set if a cached previous user id exists, it always overrides any new user id entered.

To fix this, a method was added to retrieve the userId from the payload. If a userId exists in the payload, the cached user id is not used. If a userId is not submitted in the payload (userId is null or empty), the cached userId is used to set `builder.userId`


aliasunder:
- [ ] Update tests

yuzeh:
I'm seeing this issue too. @aliasunder did you manage to work around this or did you have to fork the SDK?

PR #642, opened 2020-02-28
LIB-1574: Double-check that input stream is closed…

### LINKED ISSUE #588: Connection Resource Leak
We are seeing this in the logs with version 4.3.0:
W/OkHttpClient: A connection to https://api.segment.io/ was leaked. Did you forget to close a response body?

There's a resource leak somewhere in the Segment library.

There is also a community thread about this:
https://community.segment.com/t/y7lyfk/a-connection-to-httpsapi-segment-io-was-leaked

Thanks!

bsneed:
- Ensure that input stream is closed on a failure.  This appears to be a known issue with HttpURLConnection.
- Was not able to reproduce said issue, but this seems to follow community guidance in resolving the problem.
- Best attempt to fix #588

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/Client.java
+++ b/analytics/src/main/java/com/segment/analytics/Client.java
@@ -58,7 +58,9 @@ class Client {
           if (responseCode >= 300) {
             String responseBody;
             try {
-              responseBody = readFully(getInputStream(connection));
+              InputStream inputStream = getInputStream(connection);
+              responseBody = readFully(inputStream);
+              inputStream.close();
             } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


aultimus:
Were we seeing some errors which prompted this change?
When I try and navigate to https://segment.atlassian.net/browse/LIB-1547 i seem to get redirected to a SRC ticket?
Is it possible for you to link to aforementioned community guidance?
Have you tested this change?

bsneed:
The ticket number was initially transposed, fixed in the description.  Community guidance was seeing the fixes made in Github by others addressing the problem and using HttpURLConnection in a similar way.  Unit tests cover this change.

PR #649, opened 2020-04-01
Reproduce Lifecyce issue

Sloy:
This PR reproduces an issue with lifecycle detection when initializing the Segment SDK lazily (i.e. in an Activity instead of the Application object).

To summarize, the first `Application Opened` event is never triggered. After that, **every time** a new Activity is opened we see Segment sends an `Application Backgrounded` and an `Application Opened` event.

We opened a support ticket with more details: https://segment.zendesk.com/hc/requests/376494

DO NOT MERGE

### DIFF
--- a/analytics-samples/analytics-sample/build.gradle
+++ b/analytics-samples/analytics-sample/build.gradle
@@ -32,6 +32,9 @@ dependencies {
   implementation project(':analytics')
   implementation project(':analytics-wear')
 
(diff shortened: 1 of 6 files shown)


prayansh:
@Sloy thanks for the PR, really helpful in tackling the issue.

Sloy:
I'm glad it helped :)
And thank you for the actual fix!

PR #618, opened 2019-07-05
Migrate to AndroidX

mbernier85:
* First I updated to latest android sdk version 28
* Updated Java to compile with version 8 bytecode
* Then run migration to AndroidX

The goal is to allow user of the library to disable Jetifier

The only library causing issues here was the Annotation lib

I also updated Butterknife to 10, so it can use AndroidX

Some test lib were also updated in the migration

PR #658, opened 2020-05-10
Update project to AndroidX

lordcodes:
- Update to use AndroidX instead of the support library.
- Wear sample needed to use AndroidX Wear which is minimum API 23, however, this is only for the sample, it doesn't affect the actual wear library dependency.
- Update test/sample dependencies to latest versions.
- Update to latest Gradle version, Android Studio 3.6 Gradle plugin and compile using Android API 29.

Samples tested, all tests ran locally. Circle CI checks don't seem to appearing in the PR, so could a project maintainer let me know if I need to do anything for that.

There is already another PR (#618) to update to AndroidX, however, it has gone stale for a while.


analytics-samples/analytics-sample/src/main/java/com/segment/analytics/sample/SampleApp.java
```diff
@@ -37,10 +39,14 @@
   public void onCreate() {
     super.onCreate();
 
-    CalligraphyConfig.initDefault(
-        new CalligraphyConfig.Builder()
-            .setDefaultFontPath("fonts/CircularStd-Book.otf")
-            .setFontAttrId(R.attr.fontPath)
+    ViewPump.init(
```
chrisjenx, line 42: Oh cool, didn't know you still use this 👍

bsneed:
Wow @lordcodes; thanks so much for this!!!!

chrisjenx:
awesome thanks @lordcodes!

lordcodes:
@bsneed I mentioned in the PR description that it didn't look like the PR was running CI properly. It looks liked Circle CI only ran after you merged the PR, which failed due to an issue with test coverage.

Is having test coverage enabled crucial for you? It looks like there are issues with it, I have tried investigating why. It looks like the JaCoCo plugin the project uses has a task to generate the report that is failing:

```
Execution failed for task ':analytics:createDebugAndroidTestCoverageReport'.
> java.io.IOException: No coverage data to process in directories [PROJECT_DIR/analytics/build/outputs/code_coverage/debugAndroidTest/connected]
```

Issue #657, opened 2020-05-05
using trackDeepLinks() Validate if the intent has data to avoid an NPE


pjcyber:
The SDK is assuming if the activity has an intent it must have an URL associated. with is a wrong assumption considering activities can be launched using intent for simple data.

https://github.com/segmentio/analytics-android/blob/a9fefda981bd67e4b151f1e466e3113fdb866f00/analytics/src/main/java/com/segment/analytics/AnalyticsActivityLifecycleCallbacks.java#L101

prayansh:
@pjcyber #663 should fix this issue. Thanks for reporting

PR #663, opened 2020-05-28
Catch null intent data

niallzato:
**What does this PR do?**
When trackdeeplinks is enabled, this can result in a null pointer exception as getData() can return null while intent does not. 

**Any background context you want to provide?**
Fix for following issue: https://github.com/segmentio/analytics-android/issues/657

### DIFF
--- a/analytics/src/test/java/com/segment/analytics/AnalyticsTest.java
+++ b/analytics/src/test/java/com/segment/analytics/AnalyticsTest.java
@@ -1181,6 +1181,72 @@ public class AnalyticsTest {
   }
 
   @Test
+  public void trackDeepLinks_nullData() {
+    Analytics.INSTANCES.clear();
+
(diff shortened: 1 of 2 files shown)


analytics/src/test/java/com/segment/analytics/AnalyticsTest.java
```diff
@@ -1180,6 +1180,72 @@ protected boolean matchesSafely(TrackPayload payload) {
...
+    Activity activity = mock(Activity.class);
+
+    Intent intent = mock(Intent.class);
+
+    when(activity.getIntent()).thenReturn(intent);
+    when(intent.getData()).thenReturn(null);
+
+    callback.get().onActivityCreated(activity, new Bundle());
+
+    verify(integration, never())
```
prayansh, line 1238: is the intention here to check if no interactions have occurred with the `integration` object or just this specific interaction?

prayansh:
closes #657

PR #664, opened 2020-05-28
Add support for Destination Middleware

prayansh:
## Summary
This PR implements destination specific middleware that function similar to source middleware, but instead run only for the specified integration.
These can be configured via the `Analytics.Builder`.
Example
```java
builder
    .useDestinationMiddleware("Segment.io",
      new Middleware() {
          @Override
          public void intercept(Chain chain) {
            // ...transform payload
            chain.proceed(chain.payload()); // send new payload to the chain
          }
    )
```

This PR also deprecates `middleware` for `sourceMiddleware` which are middleware that run for all integrations.

Ref: LIBMOBILE-38

## Test Cases
- single destination middleware runs (`DestinationMiddlewareTest.middlewareRuns`)
- destination middleware only runs for specified destination (`DestinationMiddlewareTest.middlewareDoesNotRunForOtherIntegration`)
- same middleware can be used for multiple integrations (`DestinationMiddlewareTest.middlewareWillRunForMultipleIntegrations`)
- Middleware can drop event for specified integration
(`DestinationMiddlewareTest.middlewareCanShortCircuit`)
- Middleware can chain
(`DestinationMiddlewareTest.middlewareCanChain`)
- Middleware can transform the payload
(`DestinationMiddlewareTest.middlewareCanTransform`)


analytics/src/main/java/com/segment/analytics/IntegrationOperation.java
```diff
@@ -154,141 +159,159 @@ public String toString() {
...
+    Middleware.Chain chain = new MiddlewareChainRunner(0, payload, middleware, callback);
+    chain.proceed(payload);
   }
 
-  static IntegrationOperation group(final GroupPayload groupPayload) {
+  /**
+   * Integration Operation for a Segment Event (track | identify | alias | group | screen) Operation
+   * runs destination middleware for given integration before sending to the desired integration
+   */
+  static IntegrationOperation segmentEvent(
```
prayansh, line 184: cant think of a better name 😓

analytics/src/main/java/com/segment/analytics/internal/Utils.java
```diff
@@ -235,6 +235,7 @@ public static String assertNotNullOrEmpty(String text, @Nullable String name) {
   /** Returns an immutable copy of the provided map. */
   @NonNull
   public static <K, V> Map<K, V> immutableCopyOf(@NonNull Map<K, V> map) {
+    // TODO (major version change) change this out to allow @Nullable input and guard against it
```
prayansh, line 238: ideally this would become 
```
@NonNull
  public static <K, V> Map<K, V> immutableCopyOf(@Nullable Map<K, V> map) {
    if (isNullOrEmpty(map)) {
      return Collections.emptyMap();
    }
    return Collections.unmodifiableMap(new LinkedHashMap<>(map));
  }
```
but dont wanna do a breaking change in this version.

analytics/src/test/java/com/segment/analytics/IntegrationOperationTest.java
```diff
@@ -361,4 +366,44 @@ public void eventPlanOverridesSchemaDefault() throws IOException {
                 + "}"));
     verify(integration, never()).track(payload);
   }
+
+  @Test
```
prayansh, line 370: These tests have nothing to do with destination middleware, but felt good to have as unit tests

analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -1255,16 +1253,51 @@ public Builder trackDeepLinks() {
...
+     */
     public Builder middleware(Middleware middleware) {
+      return useSourceMiddleware(middleware);
+    }
+
+    /**
+     * Add a {@link Middleware} custom source middleware. This will be run before sending to all
+     * integrations
+     */
+    public Builder useSourceMiddleware(Middleware middleware) {
```
prayansh, line 1268: drop-in replacement for `middleware`

bsneed: *perfect*

analytics/src/main/java/com/segment/analytics/IntegrationOperation.java
```diff
@@ -154,141 +159,159 @@ public String toString() {
...
       @Override
-      public void run(String key, Integration<?> integration, ProjectSettings projectSettings) {
-        if (isIntegrationEnabled(groupPayload.integrations(), key)) {
-          integration.group(groupPayload);
-        }
+      void run(String key, Integration<?> integration, ProjectSettings projectSettings) {
+        List<Middleware> applicableMiddleware = getMiddlewareList(destinationMiddleware, key);
+        runMiddlewareChain(
+            payload,
+            applicableMiddleware,
```
bsneed, line 192: the isIntegrationEnabled bit that seems to have been removed here seems like it might be important.  If an integration is disabled on segment.com, there's no need to run it's middleware right?

bsneed: never mind, I see it down below.  however, it's repeated.  seems like that could be placed here instead and checked once?

analytics/src/main/java/com/segment/analytics/internal/Utils.java
bsneed: 👍

bsneed: Might be good to leave that in the TODO comment in the code itself as well.

analytics/src/main/java/com/segment/analytics/IntegrationOperation.java
bsneed: I don't have any better ideas. :(

PR #667, opened 2020-06-05
Use AndroidX lifecycle functions for App Fore/Backgrounded

prayansh:
## Summary
Use `androidx.lifecycle` methods to fire `Application Backgrounded | Application Foregrounded` track calls when `trackApplicationLifecycleEvents=true`.
This feature allows us to load Segment Analytics "lazily" via an Activity rather than via the Application and still hook into all the lifecycle methods. This is due to the fact that `AnalyticsActivityLifecycleCallbacks.onCreate` gets called even if Analytics is loaded after the `Application.onCreate`.
The Lifecycle methods are also the recommended method of determining if an application is foregrounded/backgrounded.

## Test Plan
- `Application Backgrounded | Application Foregrounded` are sent when app starts. (Segment initialized via Application)
- `Application Backgrounded | Application Foregrounded` are sent when app starts. (Segment initialized via Activity)
- `Application Installed` are sent when app starts. (Segment initialized via Application)
- `Application Installed` are sent when app starts. (Segment initialized via Activity)

### DIFF
--- a/analytics/build.gradle
+++ b/analytics/build.gradle
@@ -6,7 +6,6 @@ apply from: rootProject.file('gradle/android.gradle')
 dependencies {
   api rootProject.ext.deps.supportAnnotations
 
-//  implementation 'androidx.lifecycle:lifecycle-extensions:2.2.0'
   implementation 'androidx.lifecycle:lifecycle-process:2.2.0'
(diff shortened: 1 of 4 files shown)


prayansh:
Closes #649

PR #673, opened 2020-07-22
Converted LoggerTest to Kotlin

Jessica-Kuo:
Fixed styling on some functions


analytics/src/test/java/com/segment/analytics/LoggerTest.kt
```diff
@@ -0,0 +1,128 @@
...
+        Assertions.assertThat(ShadowLog.getLogs())
+                .containsExactly(
+                       LogItemBuilder()
+                                .type(Log.VERBOSE)
+                                .msg("some message with an argument")
+                                .build())
+    }
+
+    @Test
+    fun debugMessagesShowInLog(){
```
migs647, line 40: Nitpick: Try and keep your style and spacing consistent. In this case there should be a space between ) and {

`fun debugMessagesShowInLog() {`

analytics/src/test/java/com/segment/analytics/LoggerTest.kt
```diff
@@ -0,0 +1,128 @@
...
+        Assertions.assertThat(ShadowLog.getLogs())
+                .containsExactly(
+                LogItemBuilder()
+                        .type(Log.DEBUG)
+                        .msg("some message with an argument")
+                        .build())
+    }
+
+    @Test
+    fun infoMessagesShowInLog(){
```
migs647, line 54: Same nit as above

analytics/src/test/java/com/segment/analytics/LoggerTest.kt
```diff
@@ -0,0 +1,128 @@
...
+                .containsExactly(
+                LogItemBuilder()
+                        .type(Log.INFO)
+                        .msg("some message with an argument")
+                        .build())
+    }
+
+    @Test
+    @Throws
+    fun errorMessagesShowInLog(){
```
migs647, line 69: Same nit as above

analytics/src/test/java/com/segment/analytics/LoggerTest.kt
```diff
@@ -0,0 +1,128 @@
...
+                LogItemBuilder()
+                        .type(Log.ERROR)
+                        .throwable(throwable)
+                        .msg("some message with an argument")
+                        .build())
+    }
+
+    @Test
+    @Throws
+    fun subLog(){
```
migs647, line 85: Same nit as above

analytics/src/test/java/com/segment/analytics/LoggerTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+        Assertions.assertThat(ShadowLog.getLogs()).containsExactly(
+                LogItemBuilder()
+                        .tag("Analytics-foo")
+                        .type(Log.DEBUG)
+                        .msg("some message with an argument")
+                        .build())
+    }
+
+    class LogItemBuilder{
```
migs647, line 98: Same nit as above. `class LogItemBuilder {`

analytics/build.gradle
```diff
@@ -30,7 +31,12 @@ dependencies {
   // with Android.
   //noinspection GradleDependency
   testImplementation 'org.assertj:assertj-core:1.7.1'
+  implementation "androidx.core:core-ktx:+"
```
egor-n, line 34: Should the dependencies be pinned to a specific version instead? This way there are no surprises when the API changes in the new updates.

Also, it seems like this dependency isn't actually required right now.

PR #674, opened 2020-07-23
Converted PropertiesTest and TraitsTest to Kotlin

Jessica-Kuo:


prayansh:
Since we r migrating tests, one good thing to add here would be screenshots showing that coverage hasnt decreased. Maybe like a before and after state.

Jessica-Kuo:
Right, that would be a quick way to show the tests produce the same results. I'll get screenshots of the test results from running the tests in Java and then running them with these files.

migs647:
Also, lets create a new Pull Request so we don't keep adding changes building the changes up that have already been merged.

build.gradle
```diff
@@ -1,5 +1,6 @@
 buildscript {
-  repositories {
+    ext.kotlin_version = '1.3.72'
```
prayansh, line 2: can we move this here? https://github.com/segmentio/analytics-android/blob/master/build.gradle#L47

would make sense to have all the configs together

PR #675, opened 2020-07-24
converted properties and traits test to kotlin

Jessica-Kuo:


Jessica-Kuo:
propertiesTest kt-coverage
traitsTest kt-coverage

I had deleted the java files before saving their coverage reports but I'll be sure to save coverage reports of future tests!

Jessica-Kuo:
OptionsTest.java coverage:
optionsTest java-coverage

OptionsTest.kt coverage:
optionsTest kt-coverage

StatsTest.java coverage:
statsTest kt-coverage

StatsTest.kt coverage:
statsTest java-coverage

prayansh:
i should have been more clear 😅 . Can u screenshot the test coverage for the individual files that these tests are unit testing. So for example: OptionsTest.kt -> show Options.java coverage.

Jessica-Kuo:
StatsTest.java:
StatsTest-j

StatsTest.kt:
StatsTest-kotlin

OptionsTest.java:
OptionsTest-j

OptionsTest.kt:
OptionsTest-kotlin

analytics/src/test/java/com/segment/analytics/PropertiesTest.kt
```diff
@@ -0,0 +1,179 @@
...
+    var properties: Properties? = null
+
+    @Before
+    fun setUp() {
+        properties = Properties()
+    }
+
+    @Test
+    fun revenue() {
+        properties?.putRevenue(3.14)
```
prayansh, line 18: is there a way to get rid of these `?` since we know in this test scenario that Properties is never null?

analytics/src/test/java/com/segment/analytics/OptionsTest.kt
```diff
@@ -0,0 +1,71 @@
...
+        } catch(expected: IllegalArgumentException){
+            Assertions.assertThat(expected)
+                    .hasMessage("Segment integration cannot be enabled or disabled.")
+        }
+    }
+
+    @Throws(Exception::class)
+    @Test
+    fun setIntegration() {
+        options?.setIntegration("Mixpanel", true)
```
prayansh, line 36: same comment here

analytics/src/test/java/com/segment/analytics/StatsTest.kt
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import java.io.IOException
+
+@RunWith(RobolectricTestRunner::class)
+@Config(manifest = Config.NONE)
+class StatsTest {
+    private var stats: Stats? = null
```
prayansh, line 16: same comment here. Is there a way to use `Stats` instead of `Stats?`. In these test scenarios we can confirm that this object is never null so this verbosity is kinda off-putting

PR #681, opened 2020-07-30
Make new Lifecycle flow opt-out

prayansh:
This is potential fix for issue currently faced by some android tv users. 
This change doesnt completely remove the new lifecycle process, but rather allows switching between the new and legacy flow. 
A new builder option `useNewLifecycleMethods` is introduced which defaults to true, and to opt-out of the new flow use
```java
Analytics.Builder builder =
        new Analytics.Builder(this, ANALYTICS_WRITE_KEY)
            ...
            .useNewLifecycleMethods(false)
```


analytics/src/main/java/com/segment/analytics/AnalyticsActivityLifecycleCallbacks.java
```diff
@@ -50,14 +52,46 @@
   private AtomicInteger numberOfActivities;
   private AtomicBoolean firstLaunch;
 
+  private AtomicBoolean isChangingActivityConfigurations;
+  private Boolean useNewLifecycleMethods;
+
+  private static LifecycleOwner stubOwner =
```
prayansh, line 58: This is basically a stub which is only used when we dont use the new lifecycle flow

prayansh: required bcos some of the functions take in `NonNull` params, so this gets rid of those warnings

prayansh:
Implementation:
This code path controls the use of the new lifecycle flow: https://github.com/segmentio/analytics-android/pull/681/files#diff-a18a86eda368939313a371b220e906f0R331
If we circumvent ^ registration, then the onStart, onCreate, onStop functions in AnalyticsActivityLifecycleCallbacks wont get implicitly called, so we add these changes
https://github.com/segmentio/analytics-android/pull/681/files#diff-12bb80c3903ca5bd6c71fc44ea4ad80bR211
https://github.com/segmentio/analytics-android/pull/681/files#diff-12bb80c3903ca5bd6c71fc44ea4ad80bR197
https://github.com/segmentio/analytics-android/pull/681/files#diff-12bb80c3903ca5bd6c71fc44ea4ad80bR158
that will call the functions via the legacy lifecycle flow

PR #693, opened 2020-08-05
converted AnalyticsBuilderTest to Kotlin

Jessica-Kuo:
converted AnalyticsBuilderTest to Kotlin

AnalyticsBuilderTest.java coverage:
AnalyticsBuilder-java

AnalyticsBuilderTest.kt coverage:
AnalyticsBuilder-kotlin


analytics/src/test/java/com/segment/analytics/AnalyticsBuilderTest.kt
```diff
@@ -0,0 +1,280 @@
...
+class AnalyticsBuilderTest {
+
+    lateinit var context: Application
+
+    @Before
+    fun setUp() {
+        MockitoAnnotations.initMocks(this)
+        context = TestUtils.mockApplication()
+        Analytics.INSTANCES.clear()
+        Mockito.`when`(context.applicationContext).thenReturn(context)
```
prayansh, line 33: why is the `when` wrapped in backticks?

prayansh: oh i see! its cos `when` is a keyword in kotlin.
MIght wanna try this out then
https://www.baeldung.com/kotlin-mockito 
```
library exposes a new whenever() method that can be used freely, without the need for back-ticks like we had to when using Mockito's native when() method.
```

analytics/build.gradle
```diff
@@ -31,6 +33,10 @@ dependencies {
   // with Android.
   //noinspection GradleDependency
   testImplementation 'org.assertj:assertj-core:1.7.1'
+  testImplementation group: 'net.bytebuddy', name: 'byte-buddy', version: '1.9.0'
+  testImplementation group: 'net.bytebuddy', name: 'byte-buddy-agent', version: '1.9.0'
+  //testImplementation 'org.objenesis:objenesis:2.6'
```
migs647, line 38: We should remove this commented out library

PR #702, opened 2020-09-10
Edge Function Middleware API

migs647:
Updates to analytics API


analytics-samples/analytics-sample/build.gradle
```diff
@@ -46,4 +46,5 @@ dependencies {
   androidTestImplementation 'com.segment.backo:backo:1.0.0'
 
   androidTestImplementation 'org.assertj:assertj-core:3.16.1'
+  implementation project(':jsmiddleware')
```
migs647, line 49: This needs to be removed

analytics-samples/analytics-sample/src/main/java/com/segment/analytics/sample/SampleApp.java
```diff
@@ -25,82 +25,99 @@
...
 import io.github.inflationx.calligraphy3.CalligraphyConfig;
 import io.github.inflationx.calligraphy3.CalligraphyInterceptor;
 import io.github.inflationx.viewpump.ViewPump;
 
 public class SampleApp extends Application {
 
-  // https://segment.com/segment-engineering/sources/android-test/settings/keys
-  private static final String ANALYTICS_WRITE_KEY = "HO63Z36e0Ufa8AAgbjDomDuKxFuUICqI";
+    // https://segment.com/segment-engineering/sources/android-test/settings/keys
+    private static final String ANALYTICS_WRITE_KEY = "1vCwH2CEddxqO0nsk7rKezv0gzFNlJGH";
```
migs647, line 43: Should we reset this write key?

migs647: Maybe we should do something generic key with a TODO above it for when people want to use the sample app they'll know where to add their own key easily via the IDE

analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -1310,6 +1315,16 @@ public Builder useDestinationMiddleware(String key, Middleware middleware) {
       return this;
     }
 
+    public Builder useEdgeFunctionMiddleware(JSMiddleware middleware) {
+
```
prayansh, line 1319: We should add an IllegalStateException here, in the case that sourceMiddleware/destinationMiddleware are already used

analytics/src/main/java/com/segment/analytics/ConnectionFactory.java
```diff
@@ -24,7 +24,7 @@
 package com.segment.analytics;
 
 import android.util.Base64;
-import com.segment.analytics.core.BuildConfig;
+import com.segment.analytics.core.*;
```
prayansh, line 27: what is this change for?

build.gradle
```diff
@@ -1,5 +1,6 @@
 buildscript {
-  ext.kotlin_version = '1.3.72'
+    ext.kotlin_version = '1.4.0'
+    ext.kotlin_version = '1.3.72'
```
prayansh, line 3: lets delete this line

analytics-samples/analytics-sample/build.gradle
prayansh, line 49: lets delete this line

settings.gradle
```diff
@@ -3,3 +3,5 @@ include 'analytics-tests'
 include 'analytics-wear'
 include 'analytics-samples:analytics-sample'
 include 'analytics-samples:analytics-wear-sample'
+include 'jsruntime'
+include 'jsmiddleware'
```
prayansh, line 7: lets delete these  lines

analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -1398,11 +1413,20 @@ public Analytics build() {
...
               ? Collections.<String, List<Middleware>>emptyMap()
               : immutableCopyOf(this.destinationMiddleware);
+      if (this.edgeFunctionMiddleware != null
+          && this.edgeFunctionMiddleware.sourceMiddleware != null) {
+        srcMiddleware = this.edgeFunctionMiddleware.sourceMiddleware;
+      }
+      if (this.edgeFunctionMiddleware != null
+          && this.edgeFunctionMiddleware.destinationMiddleware != null) {
+        destMiddleware = this.edgeFunctionMiddleware.destinationMiddleware;
+      }
```
prayansh, line 1429: I think this could raise a scenario where edge function is inter-operating with either native source/destination middleware. 
Lets try to have parity with iOS here

prayansh (commented):
left some comments

prayansh (changes requested):
Lets also add some tests for the new behaviours

analytics/src/test/java/com/segment/analytics/EdgeFunctionMiddlewareTest.kt
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Test
+    @Throws(Exception::class)
+    fun edgeFunctionMiddlewareOverwrites() {
+
+        try {
+            val analytics = builder
+                    .useEdgeFunctionMiddleware(Mockito.mock(JSMiddleware::class.java))
+                    .useSourceMiddleware(
```
prayansh, line 96: just wanna confirm that we are ok with letting users specify both without raising an exception in this scenario

analytics/src/test/java/com/segment/analytics/EdgeFunctionMiddlewareTest.kt
```diff
@@ -0,0 +1,141 @@
...
+                            Middleware { throw AssertionError("should not be invoked") })
+                    .useSourceMiddleware(
+                            Middleware { throw AssertionError("should not be invoked") })
+                    .build()
+
+            Assertions.fail("Should not reach this state")
+        } catch (exception: java.lang.Exception) {
+            Assertions.assertThat(exception).isInstanceOf(IllegalStateException::class.java)
+        }
+    }
```
prayansh, line 140: can we also add a test for the scenario where we add a source/destination middleware before edge functions and an exception is raised?

prayansh (approved):
LGTM

PR #707, opened 2020-09-17
Remove the ability to natively report attribution information via Segment integrations

prayansh:
This deprecates the `trackAttributionInformation` builder option. Using it will result in a no-op. 
This PR also removes all interactions with `mobile-service` and any codepath that depends on the same.


bsneed (approved):

analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -1283,9 +1239,12 @@ public Builder recordScreenViews() {
             return this;
         }
 
-        /** Automatically track attribution information from enabled providers. */
+        /**
+         * Automatically track attribution information from enabled providers. This build option has
```
prayansh, line 1243: any recommendations on the text here??

PR #708, opened 2020-09-17
Kotlin/analytics test conversion

Jessica-Kuo:
AnalyticsTest.java coverage:
AT java coverage

AnalyticsTest.kt coverage:
AT kt coverage


analytics/src/test/java/com/segment/analytics/EdgeFunctionMiddlewareTest.kt
```diff
@@ -108,7 +108,7 @@ class EdgeFunctionMiddlewareTest {
         analytics.track("foo")
 
         val privateEdgeFunctions: Field = Analytics::class.java.getDeclaredField("edgeFunctionMiddleware")
-        Assertions.assertThat(privateEdgeFunctions).isNotNull
+        Assertions.assertThat(privateEdgeFunctions).isNotNull()
```
prayansh, line 111: good catch :)

Jessica-Kuo: Yay! I wanted to look at the new test and then asked Cody what to do when the error was there.

prayansh: surprising how the CI was passing without this 🤔

migs647: Yah they pass locally too without the parens, that's why I didn't catch it, but gives some weird non-compile warning.

prayansh (approved):
LGTM 🚢

PR #711, opened 2020-09-22
Kotlin/segment integration test

Jessica-Kuo:
SegmentIntegrationTest.java coverage:
SIT-J

SegmentIntegrationTest.kt coverage:
SIT-K


analytics/src/test/java/com/segment/analytics/SegmentIntegrationTest.kt
```diff
@@ -0,0 +1,653 @@
...
+                .messageId("a161304c-498c-4830-9291-fcfb8498877b")
+                .timestamp(parseISO8601Date("2014-12-15T13:32:44-0700"))
+                .event("foo")
+                .userId("userId")
+                .build()
+
+        segmentIntegration.performEnqueue(trackPayload)
+        val expected =
+            (
+                "{" +
```
prayansh, line 147: we should leverage kotlin's multiline string here

analytics/src/test/java/com/segment/analytics/SegmentIntegrationTest.kt
```diff
@@ -0,0 +1,653 @@
...
+    }
+
+    @Test
+    @Throws(IOException::class)
+    fun payloadVisitorReadsOnly475KB() {
+        val payloadWriter = PayloadWriter(
+            mock(BatchPayloadWriter::class.java), Crypto.none()
+        )
+        val bytes =
+            """{
```
prayansh, line 476: for readability it would be good if the starting `{` is on a different line. something like
```
"""
{
   ....
}
"""

**Review by prayansh — CHANGES_REQUESTED** [review]
just some minor suggestions to improve readability/convention

File: analytics/src/test/java/com/segment/analytics/SegmentIntegrationTest.kt [hunk 3]
```diff
@@ -0,0 +1,653 @@
...
+                    override fun close() {
+                        throw Client.HTTPException(400, "Bad Request", "bad request")
+                    }
+                })
+        val segmentIntegration =
+            SegmentBuilder()
+                .client(client)
+                .payloadQueue(payloadQueue)
+                .build()
+        for (i in 0 until 4) {
```
migs647, line 353: I kind of prefer `for (i in 0..3)` but either are okay :)

migs647 (approved):
LGTM

analytics/src/test/java/com/segment/analytics/SegmentIntegrationTest.kt
Jessica-Kuo: I tried changing the brackets to be in that form but it changes the output of the test and makes it fail? I think this is because of how byte arrays are formed?

prayansh: oh interesting. i guess ignore for now

prayansh (approved):
🚢

PR #720, opened 2020-09-29
converted TrackPayloadTest to Kotlin

Jessica-Kuo:
TrackPayloadTest.java coverage:
TPT-J

TrackPayloadTest.kt coverage:
TPT-K


analytics/src/test/java/com/segment/analytics/integrations/TrackPayloadTest.kt
```diff
@@ -0,0 +1,100 @@
...
+    lateinit var builder: TrackPayload.Builder
+
+    @Before
+    fun setUp() {
+        builder = TrackPayload.Builder().userId("userId")
+    }
+
+    @Test
+    fun invalidEventThrows() {
+        //noinspection CheckResult,ConstantConditions
```
prayansh, line 44: do we still need this noinspection comment?

prayansh (approved):
LGTM; there are a couple `noinspection` comments which may no longer be required. if thats the case then we should remove them

PR #728, opened 2020-10-15
Enable faster project settings refresh when debugging enabled

prayansh:
In order to refresh settings faster when debugging, we now change the mobile cache ttl to 1 minute when in debug mode.


migs647 (approved):
LGTM

PR #730, opened 2020-10-23
additional identify tests for LIBMOBILE-49

Jessica-Kuo:
LIBMOBILE-49: https://segment.atlassian.net/browse/LIBMOBILE-49

### Things that couldn't be tested:
analytics.identify(null) throws an IllegalArgumentException so that couldn't be tested for resetting traits (there is an existing unit test for this).
analytics.identify() cannot be tested since the command requires at least one parameter.
analytics.identify("userId") will not have null traits since userId is a trait.

### New tests:
Test identifyNullTraits makes sure that when a userId is given, other traits aren't set (tested with username)
Test identifySavesPreviousTraits makes sure that when traits are defined under a userId they carry over when an identify call with only that userId is given.

### DIFF
--- a/analytics/src/test/java/com/segment/analytics/AnalyticsTest.kt
+++ b/analytics/src/test/java/com/segment/analytics/AnalyticsTest.kt
@@ -265,6 +265,23 @@ open class AnalyticsTest {
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


migs647 (approved):
LGTM

PR #742, opened 2020-12-18
Fixed StrictMode DiskReadViolation and DiskWriteViolation

Jessica-Kuo:
File containing logcat of the StrictMode error: (starts at line 152) 
[StrictMode_Logcat_before.pdf](https://github.com/segmentio/analytics-android/files/5718129/StrictMode_Logcat_before.pdf)

File containing logcat with the fix:
[StrictMode_Logcat_after.pdf](https://github.com/segmentio/analytics-android/files/5718133/StrictMode_Logcat_after.pdf)

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/SegmentIntegration.java
+++ b/analytics/src/main/java/com/segment/analytics/SegmentIntegration.java
@@ -36,6 +36,7 @@ import android.os.HandlerThread;
 import android.os.Looper;
 import android.os.Message;
 import android.util.JsonWriter;
+import android.util.Log;
 import com.segment.analytics.integrations.AliasPayload;
 import com.segment.analytics.integrations.BasePayload;
 import com.segment.analytics.integrations.GroupPayload;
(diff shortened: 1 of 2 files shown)


prayansh:
Since the violation only happens on our own SegmentIntegration can we narrow down the scope of the thread here? I would suggest somewhere here instead: https://github.com/segmentio/analytics-android/blob/4555626794d4e63c8a1f9d98e54eca86c97f262e/analytics/src/main/java/com/segment/analytics/SegmentIntegration.java#L185

Jessica-Kuo:
Updated! New logcat file: 
[StrictMode_after_fix.pdf](https://github.com/segmentio/analytics-android/files/5718286/StrictMode_after_fix.pdf)

analytics/src/main/java/com/segment/analytics/SegmentIntegration.java
```diff
@@ -182,22 +183,36 @@ static synchronized SegmentIntegration create(
             Logger logger,
             Crypto crypto,
             ValueMap settings) {
-        PayloadQueue payloadQueue;
+        final PayloadQueue[] payloadQueue = new PayloadQueue[1];
+        Thread execThread =
+                new Thread(
+                        () -> {
+                            try {
+                                File folder =
```
prayansh, line 191: This seems like a great candidate to use Futures. Futures allow you to return values from a thread. I would suggest using those here

prayansh (changes requested):

analytics/src/main/java/com/segment/analytics/SegmentIntegration.java
prayansh: also i just wanted to confirm, but are there are any potential downfalls to creating this file from a random thread? I cant think of any but definitely something to think about

Jessica-Kuo: When I was making it, I didn't think there were any, but I also didn't look into the potential issues of isolating the process to it's own thread... from what I remember in school, as long as there isn't any other process in the thread it shouldn't deadlock (without mutexes or anything to manage it) so I think its okay?

Issue #741, opened 2020-12-14
Android 11 IllegalStateException: SharedPreferences in credential encrypted storage are not available until after user is unlocked


kozyrevsergey89:
Hi,

I'm using segment version 4.8.2 and got a crash on Android 11 (no issues on previous OS versions).

`java.lang.IllegalStateException: SharedPreferences in credential encrypted storage are not available until after user is unlocked
    at android.app.ContextImpl.getSharedPreferences(ContextImpl.java:494)
    at android.app.ContextImpl.getSharedPreferences(ContextImpl.java:479)
    at android.content.ContextWrapper.getSharedPreferences(ContextWrapper.java:188)
    at com.segment.analytics.internal.Utils.getSegmentSharedPreferences(Utils.java:309)
    at com.segment.analytics.ValueMap$Cache.<init>(ValueMap.java:395)
    at com.segment.analytics.ProjectSettings$Cache.<init>(ProjectSettings.java:76)
    at com.segment.analytics.Analytics$Builder.build(Analytics.java:1378)`

This happens when we schedule some work with WorkManager and it executes after device is booted but user hasn't entered credentials for the first time. Job execution can not be moved to a later stage, and we still need to send analytics.

kozyrevsergey89:
there is workaround suggested by Google: https://github.com/googlearchive/android-DirectBoot/blob/737ab9a214debbf3c940360b834077ccc3035ff3/Application/src/main/java/com/example/android/directboot/alarms/AlarmStorage.java

aloz77:
Still running into this issue with ACRA 5.11.3 when the Application is staring on LOCKED_BOOT_COMPLETED and calling ACRA.init(). If there is no general solution, is there a chance to add a method to ACRA.init() to set the context for the preference storage manually?

PR #743, opened 2021-01-04
fix java.lang.IllegalStateException: SharedPreferences in credential encrypted storage are not available until after user is unlocked by using device protected storage for segment shared preferences

kozyrevsergey89:
fix java.lang.IllegalStateException: SharedPreferences in credential encrypted storage are not available until after user is unlocked by using device protected storage for segment shared preferences.
[Issue tracked her.](https://github.com/segmentio/analytics-android/issues/741)

The issue reproduces when we try to send any analytics with segment lib from a scheduled job or from push notification handling while device is in booted but locked state. Meaning, user turned on the device, but hasn't unlock the device at least once. At this moment only device protected storage is available, but default storage that is used in current implementation - credentials protected storage which is only available when user unlocks the device.

The fix is pretty much adapted Google solution they propose in their [directBoot example app](https://github.com/googlearchive/android-DirectBoot/blob/737ab9a214debbf3c940360b834077ccc3035ff3/Application/src/main/java/com/example/android/directboot/alarms/AlarmStorage.java).

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/internal/Utils.java
+++ b/analytics/src/main/java/com/segment/analytics/internal/Utils.java
@@ -45,6 +45,10 @@ import android.telephony.TelephonyManager;
 import android.text.TextUtils;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+
+import com.segment.analytics.Analytics;
+import com.segment.analytics.integrations.Logger;
+
 import java.io.BufferedReader;
 import java.io.Closeable;
 import java.io.File;
@@ -81,6 +85,7 @@ public final class Utils {
     public static final int DEFAULT_FLUSH_QUEUE_SIZE = 20;
     public static final boolean DEFAULT_COLLECT_DEVICE_ID = true;
     public static final String DEFAULT_API_HOST = "api.segment.io/v1";
+    private static final Logger logger = Logger.with(Analytics.LogLevel.DEBUG);
 
     /** Creates a mutable HashSet instance containing the given elements in unspecified order */
     public static <T> Set<T> newSet(T... values) {
@@ -308,8 +313,21 @@ public final class Utils {
 
     /** Returns a shared preferences for storing any library preferences. */
     public static SharedPreferences getSegmentSharedPreferences(Context context, String tag) {
-        return obtainDeviceProtectedStorageContext(context)
-                .getSharedPreferences("analytics-android-" + tag, MODE_PRIVATE);
+        Context storageContext;
+        String prefsName = "analytics-android-" + tag;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            // All N devices have split storage areas, but we may need to recreate the preferences in the new device
+            // protected storage area, which is where the data lives from now on.
+            final Context deviceContext = context.createDeviceProtectedStorageContext();
+            if (!deviceContext.moveSharedPreferencesFrom(context,
+                    prefsName)) {
+                logger.debug("Failed to migrate shared preferences.");
+            }
+            storageContext = deviceContext;
+        } else {
+            storageContext = context;
+        }
+        return storageContext.getSharedPreferences(prefsName, MODE_PRIVATE);
     }
 
     /** Get the string resource for the given key. Returns null if not found. */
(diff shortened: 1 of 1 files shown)


prayansh (approved):

prayansh:
Thanks for the PR @kozyrevsergey89 :)

prayansh:
@kozyrevsergey89 is it possible to write a test that covers this scenario? Would be great to have that coverage 😄

kozyrevsergey89:
No you can only manually test it. I'll write step by step

analytics/src/main/java/com/segment/analytics/internal/Utils.java
```diff
@@ -532,6 +533,16 @@ public static void copySharedPreferences(SharedPreferences src, SharedPreference
         editor.apply();
     }
 
+    private static Context obtainDeviceProtectedStorageContext(Context context) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            // All N devices have split storage areas, but we may need to recreate the preferences in the new device
+            // protected storage area, which is where the data lives from now on.
+            return context.createDeviceProtectedStorageContext();
```
artour-bakiev, line 540: Would it be a backward-compatibility issue? Not 100% sure but probably it's the client job to specify which context should be used to access the storage area - device or user...

Perhaps we can do better - we can extend the `Builder` [class](https://github.com/segmentio/analytics-android/blob/master/analytics/src/main/java/com/segment/analytics/Analytics.java#L1061)

prayansh: good callout. once we get the test plan i will try to run it via multiple scenarios to make sure nothing breaks 👍

kozyrevsergey89: I don't think it is client responsibility think of which context can be specified, given that any needed check can be done on the library side and also any type of context can be obtained.
Context class itself has those api to get application context or to get context with device protected storage reference. It just doesn't make sense to burden clients with all those decisions.

kozyrevsergey89: I'll just move prefs to device protected storage as a migration step.

artour-bakiev: > I don't think it is client responsibility think of which context can be specified

Yeah, may be it's not the client responsibility... It mostly depends on what type of data the segment library keeps in the shared preferences.

According to https://android-developers.googleblog.com/2016/04/developing-for-direct-boot.html
> All sensitive, private information, like the full message history and a read/write access token, should still be saved in credential protected storage.

Regarding
> I'll just move prefs to device protected storage as a migration step.

This's going to be tricky since we should take into account that the credential protected storage can be unaccessible (yet).

analytics/src/main/java/com/segment/analytics/internal/Utils.java
```diff
@@ -308,8 +313,21 @@ public static String getDeviceId(Context context) {
...
     public static SharedPreferences getSegmentSharedPreferences(Context context, String tag) {
-        return obtainDeviceProtectedStorageContext(context)
-                .getSharedPreferences("analytics-android-" + tag, MODE_PRIVATE);
+        Context storageContext;
+        String prefsName = "analytics-android-" + tag;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            // All N devices have split storage areas, but we may need to recreate the preferences in the new device
+            // protected storage area, which is where the data lives from now on.
+            final Context deviceContext = context.createDeviceProtectedStorageContext();
+            if (!deviceContext.moveSharedPreferencesFrom(context,
```
artour-bakiev, line 322: Not sure if it would work properly...
I see two pretty unpleasant use cases here:
1. The device [supports FBE](https://source.android.com/security/encryption/file-based) and the device is in the `locked` state (i.e. the boot is just happened and the user hasn't used the credentials to unlock device). Then `moveSharedPreferencesFrom` would return `false` since the user credential storage is not accessible (yet)
2. The device is in the `unlocked` state (i.e. the user used the credentials to unlock device). In that case `moveSharedPreferencesFrom` would always (unconditionally) overwrite what we have in the device storage at the moment.

prayansh:
@kozyrevsergey89 any updates on the Test Plan?

prayansh (changes requested):
Requesting a test plan for the changes. Will close the PR soon if I see no traction

Issue #748, opened 2021-03-11
Crash on Android with IllegalArgumentException: INTERNET permission is required


kozyrevsergey89:
I can see `packageManager has died` line just before this crash happens in our crash reporting tool.
What I've found is that this sometimes happens when updating the app while talking to packageManager, and available fix it to repeat permission request at least once - this usually helps:
https://www.codota.com/web/assistant/code/rs/5c7cac8a2f7dae0001643522#L87
An example given from evernote android app.

prayansh:
This should be resolved via #749

PR #749, opened 2021-03-11
crashing permission check repeat fix

kozyrevsergey89:
in order to address the issue with crashing permission check I introduced another hasPermission api that can repeat the call in case an exception has been thrown while checking the permission

Addresses this issue:
https://github.com/segmentio/analytics-android/issues/748

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/internal/Utils.java
+++ b/analytics/src/main/java/com/segment/analytics/internal/Utils.java
@@ -45,6 +45,10 @@ import android.telephony.TelephonyManager;
 import android.text.TextUtils;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+
+import com.segment.analytics.Analytics;
+import com.segment.analytics.integrations.Logger;
+
 import java.io.BufferedReader;
 import java.io.Closeable;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


analytics/src/main/java/com/segment/analytics/internal/Utils.java
```diff
@@ -154,7 +161,18 @@ public static float coerceToFloat(Object value, float defaultValue) {
 
     /** Returns true if the application has the given permission. */
     public static boolean hasPermission(Context context, String permission) {
-        return context.checkCallingOrSelfPermission(permission) == PERMISSION_GRANTED;
+        return hasPermission(context, permission, 0);
+    }
+
+    private static boolean hasPermission(Context context, String permission, int repeatCount) {
+        try {
```
prayansh, line 168: is it guaranteed to always work on the second try? if so then we should probably just do function calls rather than use recursion here

kozyrevsergey89: no, it is not guaranteed to work on the second try. This is why it is beneficial to keep it parametrized for future.
This is a known solution, for example you can see:
https://github.com/evernote/android-job/blob/d9cfc62d2cd52051b424efa0dcff3574287a6171/library/src/main/java/com/evernote/android/job/util/JobUtil.java#L87
in the evernote library they do the same.
Basically, if it is failing it fails for a reason of app updating simultaneously with the request for check for permissions. Most likely second invocation will be after package manager updated its state and app update is counted.
If it doesn't fail, then we see result from first try, as we are not going into recursion in case the result is false, only when there is an exception.

prayansh: I see, thanks for the context 👍

prayansh (approved):

PR #750, opened 2021-03-11
minimise throwing in Analytics.Builder() constructor

kozyrevsergey89:
introduced NonNull annotations to prevent passing null params and throwing from constructor for those cases, also applicationContext can not be null if context is not null, updated javadoc


prayansh (approved):
🔥

PR #751, opened 2021-03-15
Remove irrelevant tests for null write-keys

prayansh:
Since #750 added `@NotNull` to write-key params, we can remove tests that test those edge cases.


bsneed (approved):

analytics/src/main/java/com/segment/analytics/internal/Utils.java
```diff
@@ -203,6 +202,16 @@ public static boolean isEmpty(@Nullable CharSequence str) {
...
     }
 
+    /**
+     * Returns true if the string is empty (once trimmed).
+     *
+     * @param str the string to be examined
+     * @return true if the string is empty (once trimmed) or 0-length
+     */
+    public static boolean isEmptyOrBlank(@NotNull CharSequence str) {
+        return str.length() == 0 || getTrimmedLength(str) == 0;
```
niallzato, line 212: Ohh the check for a NaN might make sense in the utils too. Nice!

PR #752, opened 2021-03-16
Check Double NaN and Infinity Values

niallzato:
NaN vales are being added to our JSON payload resulting in the entire payload failing. Check for these vales and now set to zero.

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/Cartographer.java
+++ b/analytics/src/main/java/com/segment/analytics/Cartographer.java
@@ -214,7 +214,7 @@ public class Cartographer {
         } else if (value instanceof Number) {
             if (value instanceof Double
                     && (Double.isNaN((Double) value) || Double.isInfinite((Double) value))) {
-                value = 0;
+                value = 0.0;
             }
             writer.value((Number) value);
         } else if (value instanceof Boolean) {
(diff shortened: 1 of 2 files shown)


analytics/src/main/java/com/segment/analytics/Cartographer.java
```diff
@@ -212,6 +212,10 @@ private static void writeValue(Object value, JsonWriter writer) throws IOExcepti
         if (value == null) {
             writer.nullValue();
         } else if (value instanceof Number) {
+            if (value instanceof Double
+                    && (Double.isNaN((Double) value) || Double.isInfinite((Double) value))) {
+                value = 0;
```
prayansh, line 217: lets make the value explicity `0.0`, to conform to a double value

analytics/src/test/java/com/segment/analytics/CartographerTest.kt
```diff
@@ -298,6 +298,27 @@ class CartographerTest {
             )
     }
 
+    @Test
+    @Throws(IOException::class)
+    fun encodesInfiniteAndNanDoubles() {
```
prayansh, line 303: 🔥

prayansh (approved):
I suggested a slight change, but overall looks 🔥

PR #755, opened 2021-03-23
Check Float NaN and Infinity

Sloy:
Hi! I saw the [fix for NaN](https://github.com/segmentio/analytics-android/pull/752) values in the latest release. Thanks!

I checked the changes and saw the fix is implemented for Double only. We actually had this problem with Floats, so I tried it and Segment is still parsing them as NaN.

I replicated the fix for Doubles, and decided to send you a PR with it. I hope it helps!


prayansh (approved):
🔥  thx for the PR

Issue #746, opened 2021-03-10
Id field on Product should deserialize product_id instead of id?


Bucimis:
https://github.com/segmentio/analytics-react-native/issues/277 (mirrored https://github.com/Appboy/appboy-segment-android/issues/25)

We have an issue where a mutual customer is expecting to be able to pass `product_id` as the Id field on product objects, but only `id` is working. I believe it's because when deserializing `Product` objects, the Android code is expecting `id` instead of `product_id` https://github.com/segmentio/analytics-android/blob/006bd94aa4ed14ae3eaf9da508f2d0653d0be8c7/analytics/src/main/java/com/segment/analytics/Properties.java#L391 (which eventually flows down to our integration https://github.com/Appboy/appboy-segment-android/blob/master/appboy-segment-integration/src/main/java/com/segment/analytics/android/integrations/appboy/AppboyIntegration.java#L263)

Is that understanding correct? I think this may only really come up with react native type integrations, where things aren't typed by integrators using the Java objects before being passed into Segment.

prayansh:
Yea this seems like an oversight, i have filed an internal Jira to map `id` and `product_id` to the same value. But until that change is shipped, i would recommend users to map both `id` and `product_id` (just like you recommended [here](https://github.com/Appboy/appboy-segment-android/issues/25#issuecomment-795839422)). I will post an update here once the change is shipped

prayansh:
did not intend to close the issue 😓

prayansh:
@Bucimis #761 should fix this issue. This change will be included in our next release

PR #761, opened 2021-03-30
Map product_id for a Product

niallzato:
As per below we do not check to see if a `product_id` is present and only map for `id`
https://github.com/segmentio/analytics-android/issues/746

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/Properties.java
+++ b/analytics/src/main/java/com/segment/analytics/Properties.java
@@ -389,6 +389,7 @@ public class Properties extends ValueMap {
     public static class Product extends ValueMap {
 
         private static final String ID_KEY = "id";
+        private static final String PRODUCT_ID_KEY = "product_id";
         private static final String SKU_KEY = "sku";
         private static final String NAME_KEY = "name";
         private static final String PRICE_KEY = "price";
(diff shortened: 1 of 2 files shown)


bsneed (approved):

analytics/src/main/java/com/segment/analytics/Properties.java
```diff
@@ -422,7 +423,7 @@ public String name() {
         }
 
         public String id() {
-            return getString(ID_KEY);
+            return getString(ID_KEY) != null ? getString(ID_KEY) : getString(PRODUCT_ID_KEY);
```
prayansh, line 426: question: do u think we should give first preference to `product_id` and then fall back to `id` or vice versa?

analytics/src/test/java/com/segment/analytics/PropertiesTest.kt
```diff
@@ -197,5 +197,9 @@ class PropertiesTest {
         assertThat(product.name()).isNull()
         product.putName("name")
         assertThat(product.name()).isEqualTo("name")
+
+        product.remove("id")
```
prayansh, line 201: 🔥

prayansh (approved):

analytics/src/main/java/com/segment/analytics/Properties.java
prayansh: sounds good

PR #781, opened 2022-01-10
address violation of retrieving unsettable device

wenxi-zeng:
* address violation of retrieving unsettable device
* fix unit test

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/internal/Utils.java
+++ b/analytics/src/main/java/com/segment/analytics/internal/Utils.java
@@ -24,15 +24,10 @@
 package com.segment.analytics.internal;
 
 import static android.Manifest.permission.ACCESS_NETWORK_STATE;
-import static android.Manifest.permission.READ_PHONE_STATE;
 import static android.content.Context.CONNECTIVITY_SERVICE;
 import static android.content.Context.MODE_PRIVATE;
-import static android.content.Context.TELEPHONY_SERVICE;
-import static android.content.pm.PackageManager.FEATURE_TELEPHONY;
 import static android.content.pm.PackageManager.PERMISSION_GRANTED;
 import static android.os.Process.THREAD_PRIORITY_BACKGROUND;
-import static android.provider.Settings.Secure.ANDROID_ID;
-import static android.provider.Settings.Secure.getString;
 
 import android.annotation.SuppressLint;
 import android.app.Activity;
@@ -45,7 +40,6 @@ import android.net.NetworkInfo;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Process;
-import android.telephony.TelephonyManager;
 import android.text.TextUtils;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -62,7 +56,6 @@ import java.net.HttpURLConnection;
 import java.security.MessageDigest;
 import java.text.ParseException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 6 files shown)


analytics/src/main/java/com/segment/analytics/internal/Utils.java
```diff
@@ -302,40 +298,58 @@ public static String assertNotNullOrEmpty(String text, @Nullable String name) {
...
         return UUID.randomUUID().toString();
     }
 
+    /**
+     * Workaround for not able to get device id on Android 10 or above using DRM API {@see
+     * https://stackoverflow.com/questions/58103580/android-10-imei-no-longer-available-on-api-29-looking-for-alternatives}
+     * {@see https://developer.android.com/training/articles/user-data-ids}
+     */
+    private static String getUniqueID() {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR2) return null;
```
prayansh, line 320: any reason to return null if we are lower than API 18?

prayansh (approved):
looks good. one comment about devices < API 18

analytics/src/main/java/com/segment/analytics/internal/Utils.java
wenxi-zeng: yes, the drm api is only available in 18+. so it's either we bump the min sdk to 18 or we do a check. and the `getDeviceID` method does a null check and returns random uuid as fallback.

chao2zhang:
👋  Hello @wenxi-zeng , do you happen to know which exact section of Google Play policy in https://support.google.com/googleplay/android-developer/answer/10144311 were we previously violating?

wenxi-zeng:
hey @chao2zhang it's under this section in the table.
```
Your app collects or links persistent device identifiers (e.g., IMEI, IMSI, SIM Serial #, etc.)
```

so IMEI, build serial, and android id are considered hardware ids, especially IMEI and build serial are unsettable ids. they are being restricted to use if SDK has min target to api 29 or above.

chao2zhang:
Thank you. Looks like this is caught by [this lint check](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/HardwareIdDetector.java?q=HardwareIdDetector)

chao2zhang:
By the way, we noticed an app startup time regression of 100ms using a vanilla setup.

andrew-polk:
I'm confused why this change didn't receive more fanfare.
One could easily consider this a breaking change, yet it didn't receive even a minor version number bump nor is it clear what the implications are from the changelog. Even this PR title is vague and misleading.

Our analytics have been relying on the context.device.id field to determine unique device counts, etc. This essentially invalidates all such data since the same device before and after this change will be seen as two different devices.
(If I've misunderstood something or there is some workaround I haven't considered, please let me know.)

I understand your hand was forced by the Android policy change, but this change should be communicated better.

analytics/src/main/java/com/segment/analytics/internal/Utils.java
```diff
@@ -321,21 +324,57 @@ public static String getDeviceId(Context context) {
...
+            wvDrm = new MediaDrm(wideVineUuid);
+            byte[] wideVineId = wvDrm.getPropertyByteArray(MediaDrm.PROPERTY_DEVICE_UNIQUE_ID);
+            MessageDigest md = MessageDigest.getInstance("SHA-256");
+            md.update(wideVineId);
+            return byteArrayToHexString(md.digest());
+        } catch (Exception e) {
+            // Inspect exception
+            return null;
+        } finally {
+            if (wvDrm == null) {
```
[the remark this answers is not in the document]
szugyi: Sorry, I see it has been fixed since

PR #791, opened 2022-02-17
fix ANR issue caused by MediaDRM api

wenxi-zeng:
* fix ANR issue caused by MediaDRM api

### DIFF
--- a/analytics/src/main/java/com/segment/analytics/AnalyticsContext.java
+++ b/analytics/src/main/java/com/segment/analytics/AnalyticsContext.java
@@ -241,7 +241,8 @@ public class AnalyticsContext extends ValueMap {
     void putDevice(boolean collectDeviceID) {
         Device device = new Device();
 
-        // use empty string to indicate device id is not yet ready
+        // use empty string to indicate device id is not yet ready.
+        // the device id will be populated async (see `attachDeviceId`)
         String identifier = collectDeviceID ? "" : traits().anonymousId();
         device.put(Device.DEVICE_ID_KEY, identifier);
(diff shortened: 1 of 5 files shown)


analytics/src/main/java/com/segment/analytics/AnalyticsContext.java
```diff
@@ -234,9 +238,11 @@ public Campaign campaign() {
     }
 
     /** Fill this instance with device info from the provided {@link Context}. */
-    void putDevice(Context context, boolean collectDeviceID) {
+    void putDevice(boolean collectDeviceID) {
         Device device = new Device();
-        String identifier = collectDeviceID ? getDeviceId() : traits().anonymousId();
+
+        // use empty string to indicate device id is not yet ready
```
prayansh, line 244: lets add a comment that deviceId will be populated async

analytics/src/main/java/com/segment/analytics/internal/Utils.java
```diff
@@ -298,25 +298,12 @@ public static String assertNotNullOrEmpty(String text, @Nullable String name) {
...
-        return UUID.randomUUID().toString();
-    }
-
     /**
      * Workaround for not able to get device id on Android 10 or above using DRM API {@see
      * https://stackoverflow.com/questions/58103580/android-10-imei-no-longer-available-on-api-29-looking-for-alternatives}
      * {@see https://developer.android.com/training/articles/user-data-ids}
      */
-    private static String getUniqueID() {
+    public static String getUniqueID() {
```
prayansh, line 306: given that this is not a safe API to use anymore, would it make sense to move it into the `GetDeviceIdTask` class?

prayansh (approved):
some comments, but LGTM

PR #816, opened 2022-10-28
client payload signing

wenxi-zeng:
* updated gradle and dependencies versions
* there are unit tests failed because of new robolectric version, will have a followup pr to fix the impacted tests. the impacted tests are marked as ignore for now
* implemented client payload signing
* added a builder option to opt-in/out client payload


analytics/src/main/java/com/segment/analytics/ConnectionFactory.java
```diff
@@ -54,9 +54,15 @@ public HttpURLConnection projectSettings(String writeKey) throws IOException {
      * Return a {@link HttpURLConnection} that writes batched payloads to {@code
      * https://api.segment.io/v1/import}.
      */
-    public HttpURLConnection upload(String apiHost, String writeKey) throws IOException {
+    public HttpURLConnection upload(String apiHost, String key, boolean useSigning) throws IOException {
         HttpURLConnection connection = openConnection(String.format("https://%s/import", apiHost));
-        connection.setRequestProperty("Authorization", authorizationHeader(writeKey));
+        if (useSigning) {
+            connection.setRequestProperty("Authorization", "SEGMENT-HMAC-SHA1");
```
prayansh, line 60: this might be incorrect, starting a thread in slack

analytics/src/main/java/com/segment/analytics/Underwriter.java
```diff
@@ -0,0 +1,41 @@
...
+            key = "segment";
+        }
+
+        try {
+            String type = "HmacSHA1";
+            mac = Mac.getInstance(type);
+            SecretKeySpec secret = new SecretKeySpec(key.getBytes(), type);
+            mac.init(secret);
+        } catch (NoSuchAlgorithmException | InvalidKeyException e) {
+            e.printStackTrace();
```
stevevls, line 25: I think the `NoSuchAlgorithmException` is pretty unlikely, but what about `InvalidKeyException`?  IIUC, if we return here and the `mac` is empty, then we'd get NPEs later when trying to send the batch.  Would we want to either 1) throw out an exception so the caller can fall back to not signing or 2) log a warning and check `mac` for `null` in the `update` and `sign` methods and fall back to no signing?

analytics/src/main/java/com/segment/analytics/Underwriter.java
```diff
@@ -0,0 +1,41 @@
...
+import java.security.NoSuchAlgorithmException;
+
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+
+public class Underwriter {
+    private Mac mac;
+
+    public Underwriter(String key) {
+        if (key == null) {
```
prayansh, line 15: maybe we should add the non-nullable annotation here rather than defaulting to a value

analytics/src/main/java/com/segment/analytics/SegmentIntegration.java
```diff
@@ -379,19 +388,17 @@ void performFlush() {
         int payloadsUploaded = 0;
         Client.Connection connection = null;
         try {
+            if (useSigning) {
```
prayansh, line 391: im still thinking on if there is a better way where we can go through batch once and compute both the payload and the signing key

prayansh (commented):
might need some changes per SDD. I would think a bit more on the batch reading and sign computation bit, cos thats the biggest performance hit

analytics/src/main/java/com/segment/analytics/ConnectionFactory.java
```diff
@@ -54,9 +54,15 @@ public HttpURLConnection projectSettings(String writeKey) throws IOException {
...
+            connection.setRequestProperty("Signature", key);
```
prayansh, line 61: also the header key should be `X-Segment-Signature`

analytics/src/main/java/com/segment/analytics/Underwriter.java
wenxi-zeng: that's a great idea. we have this implemented later

analytics/src/main/java/com/segment/analytics/Underwriter.java
wenxi-zeng: yeah. this is just for the sake of unit tests, since mocked client has the writekey as null. will change it to better way

analytics/src/main/java/com/segment/analytics/ConnectionFactory.java
wenxi-zeng: discussed in thread. will come back to this once backend reaches a decision on how to do the header

analytics/src/main/java/com/segment/analytics/SegmentIntegration.java
wenxi-zeng: discussed in slack

analytics/src/main/java/com/segment/analytics/ConnectionFactory.java
wenxi-zeng: discussed in slack

stevevls (approved):
LGTM!

wenxi-zeng:
closed for alternative solution

PR #817, opened 2022-11-30
move writekey to body

wenxi-zeng:
* updated gradle and dependencies versions
* there are unit tests failed because of new robolectric version, will have a followup pr to fix the impacted tests. the impacted tests are marked as ignore for now
* move `writeKey` from auth header to batch body
* add instanceId to context object


analytics/src/main/java/com/segment/analytics/AnalyticsContext.java
```diff
@@ -139,6 +142,7 @@ static synchronized AnalyticsContext create(
         analyticsContext.putNetwork(context);
         analyticsContext.putOs();
         analyticsContext.putScreen(context);
+        analyticsContext.putTelemetry();
```
didiergarcia, line 145: Rename to `putInstanceId()` ??

didiergarcia (approved):

analytics/src/main/java/com/segment/analytics/AnalyticsContext.java
wenxi-zeng: yah. I thought about this too. but seems like the "put" functions are grouped by functionality (network, os, app, library, etc). the `instanceId` is added specifically for telemetry. we might have more fields to support telemetry in the future. thus, thought it's better to put it into a new group

PR #824, opened 2023-03-13
fix androidx lifecycle causes analytics init in background crash

wenxi-zeng:


analytics/src/main/java/com/segment/analytics/Analytics.java
```diff
@@ -352,7 +352,12 @@ public void run() {
 
         application.registerActivityLifecycleCallbacks(activityLifecycleCallback);
         if (useNewLifecycleMethods) {
-            lifecycle.addObserver(activityLifecycleCallback);
+            // NOTE: addObserver is required to run on UI thread,
+            // though we made it compatible to run rom background thread,
```
didiergarcia, line 356: typo "rom" -> "from"

didiergarcia (approved):

